PR #46, opened 2014-11-23
Simple String#singularize util method implementation

strangeworks:


jodosha:
@strangeworks Thanks but this isn't needed for now. I'll eventually open in that case.

PR #51, opened 2015-01-15
HTML escape

jodosha:
### Feature

Escape utilities for HTML contents and attributes

HTML escape utilities
Based on OWASP research and OWASP ESAPI code

See:
https://www.owasp.org
https://www.owasp.org/index.php/Cross-site_Scripting_%28XSS%29
https://www.owasp.org/index.php/XSS_(Cross_Site_Scripting)_Prevention_Cheat_Sheet
https://www.owasp.org/index.php/ESAPI
https://github.com/ESAPI/esapi-java-legacy

Ref: https://github.com/lotus/helpers/pull/7

#### Lookup chars

It internally uses Hash instances to lookup chars.

They are implemented with this syntax `{ 0x9f => true }`, which makes them equivalent to `Set.new([0x9f])`, but the lookup is faster: https://gist.github.com/jodosha/ac5dd54416de744b9600

#### Output class

The output is a `Lotus::Utils::Escape::SafeString` instance. This is a private class that is required for two reasons:

  * Don't double escape the same string (this is for `Lotus::Helpers` compatibility)
  * Leave open the possibility to developers to mark a string as safe with an higher API (eg. `#raw` in `Lotus::View` or `Lotus::Helpers`)


stevehodgkiss:
Looks great, and easy enough to follow...

lib/lotus/utils/escape.rb
```diff
@@ -0,0 +1,487 @@
...
+    # @see https://www.owasp.org/index.php/XSS_(Cross_Site_Scripting)_Prevention_Cheat_Sheet
+    # @see https://www.owasp.org/index.php/ESAPI
+    # @see https://github.com/ESAPI/esapi-java-legacy
+    module Escape
+      # Hex base for base 10 integer conversion
+      #
+      # @since x.x.x
+      # @api private
+      #
+      # @see http://www.ruby-doc.org/Fixnum.html#method-i-to_s
```
AlfonsoUceda: Page not found, the message in the web:

`We're sorry, but that page cannot be found.`

lib/lotus/utils/escape.rb
```diff
@@ -0,0 +1,487 @@
...
+      #   <a title="<%= Lotus::Utils::Escape.html('...') %>">link</a>
+      def self.html(input)
+        result = ""
+
+        input.to_s.encode(Encoding::UTF_8).chars do |chr|
+          result << HTML_CHARS.fetch(chr, chr)
+        end
+
+        result
+      end
```
AlfonsoUceda: @jodosha don't you prefer this?

```ruby
def self.html(input)
  input.to_s.encode(Encoding::UTF_8).chars.each_with_object do |chr, result|
    result << HTML_CHARS.fetch(chr, chr)
  end
end
```

AlfonsoUceda:
:+1: @jodosha I learn a lot reading your code ;) and tomorrow I'll see the references posted above.

lib/lotus/utils/escape.rb
jodosha: @AlfonsoUceda Because `#each_with_object` is slower: https://gist.github.com/jodosha/33c6f9201eaa4e4a5f63

lib/lotus/utils/escape.rb
jodosha: @AlfonsoUceda Fixed, thanks :+1:

PR #52, opened 2015-01-22
Chain append/prepend

weppos:
This PR introduces `Chain#append` and `Chain#prepend` as discussed in lotus/controller#76.

The final version is the result of several iterations and benchmarks. I kept all the changes in the repo, along with the benchmarks, so that we can analyse the different implementations.

I'm open to feedback.

The last step is to reintroduce `Chain#add`. However, I have a question. How would you love the method to be tested?

Generally, I test delegation using a stub library. Does it make sense to "clone" all the tests we have for `prepend`? May be no. I think that using a single, generic test (like passing calling the method twice, with a bunch of different versions and checking the result) is sufficient to cover most of the cases.

### DIFF
--- a/lib/lotus/utils/callbacks.rb
+++ b/lib/lotus/utils/callbacks.rb
@@ -58,7 +58,7 @@ module Lotus
           @chain.uniq!
         end
 
-        # Preprends the given callbacks to the beginning of the chain.
+        # Prepends the given callbacks to the beginning of the chain.
         #
         # @param callbacks [Array] one or multiple callbacks to add
         # @param block [Proc] an optional block to be added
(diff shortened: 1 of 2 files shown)


jodosha:
@weppos This looks good, thank you! :+1:

Where are the benchmarks? 

Testing `#add`: not sure if `Minitest::Mock` does the trick. If not, please duplicate `#prepend` tests. I prefer to not add other development dependencies. Thank you :+1:

weppos:
> Where are the benchmarks?

I pasted the content in the git commit message. See for instance fd209ae7e87b09a262fb497e46eeceb2954d2af3

jodosha:
:heart_eyes:

joneslee85:
noice!

weppos:
@jodosha it was not complete. I still need to add the alias and test it. :smile:

weppos:
Ah, ok, you added it.

jodosha:
@weppos Yes, no problem. Thank you for this PR :+1:

PR #54, opened 2015-01-27
Deprecated Lotus::Utils::Callbacks::Chain#add in favor of #append

jodosha:
Ditto.
---

Now that we have specular API: `#append` and `#prepend` we can get rid of `#add`, which is redundant and ambiguous for the new semantic.

Ref #52


joneslee85:
+1

PR #55, opened 2015-01-28
Added `[]` as alias for get in attributes class

AlfonsoUceda:
ditto


jodosha:
@AlfonsoUceda This looks great, thank you! :+1:

PR #58, opened 2015-02-06
Removed support for MRI ruby versions lower than 2.2

AlfonsoUceda:
ditto

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,21 +3,11 @@ sudo: false
 cache: bundler
 script: 'bundle exec rake test:coverage --trace'
 rvm:
-  - 2.0.0
-  - 2.1.0
-  - 2.1.1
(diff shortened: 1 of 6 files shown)


jodosha:
@AlfonsoUceda This looks good :+1: But we should hold on until the release and keep merging master here.

teddy-ma:
Should [this description](https://github.com/lotus/utils/blob/master/lib/lotus/utils/attributes.rb#L10~#L11) to be removed since ruby 2.2 has Symbol GC?

AlfonsoUceda:
@teddy-ma supporting mri ruby2.2+ that class will dissapear

jodosha:
@teddy-ma @AlfonsoUceda Yeah, that class will no longer be used. But we won't merge those 2.2 only PRs for a while.

PR #61, opened 2015-02-11
typo

teddy-ma:


jodosha:
@teddy-ma Really good catch, thank you! :+1:

PR #63, opened 2015-02-21
Introduced Utils::String#dasherize

jodosha:
## What

`Utils::String#dasherize`

## How does it works?

```ruby
require 'lotus/utils/string'

string = Lotus::Utils::String.new 'lotus utils'
string.dasherize # => "lotus-utils"

string = Lotus::Utils::String.new 'lotus_utils'
string.dasherize # => "lotus-utils"

string = Lotus::Utils::String.new 'LotusUtils'
string.dasherize # => "lotus-utils"

# etc..
```

## Why?

This is a supporting feature for Lotus::Helpers' form helper.

```erb
<%=
  form_for(:book, routes.books_path) do
    text_field :extended_title
    # ...
  end
%>
```

That `text_field` accepts a symbol which produces HTML attributes. This new `Utils::String` functionality will be used to produce `id`:

```html
<input type="text" name="book[extended_title]" id="book-extended-title" value="">
```

### DIFF
--- a/lib/lotus/utils/string.rb
+++ b/lib/lotus/utils/string.rb
@@ -46,6 +46,12 @@ module Lotus
       # @api private
       TITLEIZE_SEPARATOR = ' '.freeze
 
+      # Separator for #dasherize
+      #
+      # @since x.x.x
+      # @api private
+      DASHERIZE_SEPARATOR = '-'.freeze
+
       # Initialize the string
       #
       # @param string [::String, Symbol] the value we want to initialize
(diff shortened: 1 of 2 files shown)


lib/lotus/utils/string.rb
```diff
@@ -110,6 +116,24 @@ def underscore
...
+      # @since x.x.x
+      #
+      # @example
+      #   require 'lotus/utils/string'
+      #
+      #   string = Lotus::Utils::String.new 'Lotus Utils'
+      #   string.dasherize # => 'lotus-utils'
+      #
+      #   string = Lotus::Utils::String.new 'lotus_utils'
+      #   string.dasherize # => 'lotus-utils'
```
joneslee85: what do you think about adding following example:

```ruby
string = Lotus::Utils::String.new 'LotusUtils'
string.dasherize # => "lotus-utils"
```

joneslee85:
except above comments, it is a solid :100:

jodosha:
@joneslee85 Thank you, I've updated it.

PR #66, opened 2015-03-10
Added blank? method for string class

AlfonsoUceda:
lotus/controller#93

Added this method to use in this PR

### DIFF
--- a/lib/lotus/utils/string.rb
+++ b/lib/lotus/utils/string.rb
@@ -69,6 +69,33 @@ module Lotus
         @string = string.to_s
       end
 
+      # Detect if a string is nil or empty
+      #
+      # @return [TrueClass, FalseClass] result checked string
+      #
+      # @since x.x.x
+      #
+      # @example nil
+      #   require 'lotus/utils/string'
+      #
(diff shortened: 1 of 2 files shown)


AlfonsoUceda:
lotus/controller#94

lib/lotus/utils/string.rb
```diff
@@ -69,6 +69,33 @@ def initialize(string)
...
+      #   string = Lotus::Utils::String.new ''
+      #   string.blank? # => true
+      #
+      # @example non empty string
+      #   require 'lotus/utils/string'
+      #
+      #   string = Lotus::Utils::String.new 'lotus'
+      #   string.blank? # => false
+      def blank?
+        @string.nil? || @string.empty?
```
joneslee85: the `nil` scenario is not possible because string is always casted to_s in constructor

joneslee85:
To me, adding `blank?` to only String class does not add much benefits. What do you think about introducing `Lotus::Kernel.blank?(object)` instead? cc @jodosha

joneslee85:
For the time being, I close this ticket.

PR #67, opened 2015-03-17
Recursively initialize Lotus::Utils::Attributes

7even:
This makes `Lotus::Utils::Attributes` initialization recursive by wrapping any inner hash in another instance of `Utils::Attributes`.

Before:

``` ruby
attrs = Lotus::Utils::Attributes.new(a: { b: 1 })
# => #<Lotus::Utils::Attributes:0x007fe0a1f3f078 @attributes={"a"=>{"b"=>1}}>
attrs[:a][:b]
# => nil
```

After:

``` ruby
attrs = Lotus::Utils::Attributes.new(a: { b: 1 })
# => #<Lotus::Utils::Attributes:0x007fe6623e2ee8 @attributes={"a"=>#<Lotus::Utils::Attributes:0x007fe6623e2dd0 @attributes={"b"=>1}>}>
attrs[:a][:b]
# => 1
```

### DIFF
--- a/lib/lotus/utils/attributes.rb
+++ b/lib/lotus/utils/attributes.rb
@@ -30,7 +30,7 @@ module Lotus
       #   attributes = Lotus::Utils::Attributes.new(a: 1, b: { 2 => [3, 4] } })
       #   attributes.to_h # => { "a" => 1, "b" => { "2" => [3, 4] } }
       def initialize(hash = {})
-        @attributes = Utils::Hash.new(hash, &nil).stringify!
+        @attributes = self.class.from_hash(hash)
       end
 
       # Get the value associated with the given attribute
@@ -111,6 +111,25 @@ module Lotus
       # @see Lotus::Utils::Hash
       def to_h
         @attributes.deep_dup
+      end
+
+      alias_method :deep_dup, :to_h
+
+      # Returns a stringified version of the hash with each nested hash wrapped
+      # in an Attributes instance
+      #
+      # @param hash [#to_h] a Hash or any object that implements #to_h
+      #
+      # @return [Lotus::Utils::Hash]
+      #
+      # @since x.x.x
+      #
+      # @see Lotus::Utils::Attributes#initialize
+      def self.from_hash(hash)
+        hash.to_h.each_with_object(Utils::Hash.new) do |(k, v), result|
+          v = Attributes.new(v) if v.is_a?(::Hash)
+          result[k.to_s] = v
+        end
       end
     end
   end
--- a/lib/lotus/utils/attributes.rb
+++ b/lib/lotus/utils/attributes.rb
@@ -113,7 +113,14 @@ module Lotus
         @attributes.deep_dup
       end
 
-      alias_method :deep_dup, :to_h
+      # Equality
+      #
+      # @return [TrueClass,FalseClass]
+      #
+      # @since x.x.x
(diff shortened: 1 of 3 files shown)


jodosha:
@7even Thank you for this PR.

The idea of recursively transform `Hash` into `Utils::Attributes` is _"too smart"_. There are cases where a `Hash` should remain a `Hash`.

**Example:** `Utils::Attributes` powers `Action::Params`. This accepts a Rack env as argument for the initializer. As you know, this is a `Hash`, if we transform it into `Utils::Attributes` we break compatibility.

The bottom line is `Utils::Attributes` != `Hash`.

This leads me to a question: what are you trying to solve? In which case this bothers you? Is it for params? Since Controller 0.4.0 we'll have [`Params#get`](https://github.com/lotus/controller/blob/master/lib/lotus/action/params.rb#L148-L192). It will provide a safe way to access nested data structure: `params.get('address.street')`.

7even:
My goal is exactly accessing the action parameters. Sorry, I didn't know about `Params#get` - it seems like a really cool replacement to all that `params[:user][:address][:street]` stuff.

On the other hand I was thinking about using `Utils::Attributes` outside Lotus with the same purpose - handling input parameters. Currently `Utils::Attributes` works as a hash with indifferent access supporting just one level of nesting so I aimed at adding infinite subhashes.

I would gladly use `Action::Params` but requiring whole `lotus/controller` just for a single class sounds like an overkill to me. Did you think about moving `Action::Params#get` functionality to `Utils::Attributes#get` and then delegating to it?

> As you know, this is a Hash, if we transform it into Utils::Attributes we break compatibility.

As I see through the code the input hash is already transformed to `Lotus::Utils::Hash` (`Attributes#initialize` [calls](https://github.com/lotus/utils/blob/2ac7259c0cbfa2ea4a22d73029729dd330817d1a/lib/lotus/utils/attributes.rb#L33) `#stringify!` on `Utils::Hash` instance which in turn [destructively transforms](https://github.com/lotus/utils/blob/2ac7259c0cbfa2ea4a22d73029729dd330817d1a/lib/lotus/utils/hash.rb#L76-L79) all subhashes). Is this okay?

jodosha:
> I would gladly use Action::Params but requiring whole lotus/controller just for a single class sounds like an overkill to me.

Are you using Lotus without the controller gem?

> Did you think about moving Action::Params#get functionality to Utils::Attributes#get and then delegating to it?

`Utils::Attributes` is going to disappear. Once we'll support only Ruby 2.2+, we'll no longer need of indifferent access, so it will be removed from `Action::Params` and `Validations`.

> As I see through the code the input hash is already transformed to Lotus::Utils::Hash (Attributes#initialize calls #stringify! on Utils::Hash instance which in turn destructively transforms all subhashes). Is this okay?

In a language like Ruby the most important thing is message passing. `Utils::Hash` has a public API that is 100% compatible with `Hash`. It will be able to respond to all the messages that a developer expects from Rack env (`Hash`).

For instance, in `Action::Params`, we send the message `#has_key?`. `Utils::Hash` is able to dispatch it, while `Utils::Attributes` isn't.

7even:
> Are you using Lotus without the controller gem?

No, I was just thinking about using `lotus-utils` in a non-lotus app. I'm not really sure this is a good idea :)

I agree with your point, thanks for the feedback.

PR #70, opened 2015-04-07
Lotus::Interactor#error to return falsey value

### LINKED ISSUE #69: Interactor: recording error shouldn't return false?
Hello,

Been playing with `Lotus::Interactor` and trying to combine `valid?` requirement (since is a private method that **must** override) with logging of `errors`, and found that `error` logging is not returning a false/falsy value to be used in combination with `valid?`.

Example:

```ruby
class PublishVideo < Struct.new(:user, :video)
  include Lotus::Interactor

  def call
    puts "publish!"
    puts "reindex!"
  end

  def valid?
    owns? && non_published?
  end

  private

  def owns?
    # trivial ownership check
    (video.user_id == user.id) or
      error "You're not owner of this video"
  end

  def non_published?
    video.draft? or
      error "Can only publish a draft video"
  end
end
```

Problem is that `valid?` returns `true`, even when `error` has been called and `errors` is not empty:

```ruby
class User < Struct.new(:id)
end

class Video < Struct.new(:user_id, :draft)
  alias_method :draft?, :draft
end

user  = User.new(1)
video = Video.new(2, false)

service = PublishVideo.new(user, video)
result  = service.call

p success: result.success?,  # => false
  errors:  result.errors     # => ["You're not owner of this video", "Can only publish a draft video"]
```

Perhaps I'm misusing Interactor in this case?

Any hint will be appreciated.

Thank you.

jodosha:
Closes #69

### DIFF
--- a/lib/lotus/interactor.rb
+++ b/lib/lotus/interactor.rb
@@ -293,6 +293,8 @@ module Lotus
     #
     # @param message [String] the error message
     #
+    # @return false
+    #
     # @since 0.3.5
     #
     # @see Lotus::Interactor#error!
@@ -336,6 +338,7 @@ module Lotus
     #   result.logger # => [:prepare_data!, :persist!, :sync!]
     def error(message)
       @_errors << message
+      false
     end
 
     # Log an error AND interrupting the flow.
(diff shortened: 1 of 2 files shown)


luislavena:
@jodosha this works as expected.

Thank you!

:heart: :heart: :heart:

PR #79, opened 2015-05-20
Make TypeError messages even more useful

francois:
I was struck by a `TypeError` with no information about what went wrong:

```
TypeError: can't convert into Integer
```

The backtrace led me nowhere as this was all framework code. The last code that belonged to me was the call to `Repository#create`. After that, it was all library code. See https://gist.github.com/francois/283f9ee23d268b69aa60 for a sample backtrace.

Now, I get a much more useful error message:

```
TypeError: can't convert #<Date: 2015-04-20 ((2457133j,0s,0n),+0s,2299161j)> into Integer
```

### DIFF
--- a/lib/lotus/utils/kernel.rb
+++ b/lib/lotus/utils/kernel.rb
@@ -140,7 +140,7 @@ module Lotus
           Set.new(::Kernel.Array(arg))
         end
       rescue NoMethodError
-        raise TypeError.new("can't convert into Set")
+        raise TypeError.new("can't convert #{inspect_type_error(arg)}into Set")
       end
 
       # Coerces the argument to be a Hash.
@@ -205,7 +205,7 @@ module Lotus
             super(arg)
           end
         rescue NoMethodError
-          raise TypeError.new "can't convert into Hash"
+          raise TypeError.new "can't convert #{inspect_type_error(arg)}into Hash"
         end
       else
         def self.Hash(arg)
@@ -217,7 +217,7 @@ module Lotus
             super(arg)
           end
         rescue ArgumentError, NoMethodError
-          raise TypeError.new "can't convert into Hash"
+          raise TypeError.new "can't convert #{inspect_type_error(arg)}into Hash"
         end
       end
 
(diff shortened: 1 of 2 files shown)


francois:
Not sure if we want to make `#inspect_type_error` private or not. Also, if the argument is a collection, it may not be a good idea to inspect it. The collection might contain thousands of elements. I'll be happy to contribute changes if necessary.

joneslee85:
@francois do you use lotus-model in your code at all?

francois:
I found my issue: I was using a composite primary key for my database table, where the first component of the primary key was a date. Sequel returns the first element of the composite primary key, instead of returning the full primary key.

I have switched my primary key around to a plain surrogate key, and all is good now.

This pull request is still valid, as it helps developers understand what value caused the error.

francois:
To answer your question, @joneslee85, I do use the latest released version of lotus-model, as specified by lotusrb.

lib/lotus/utils/kernel.rb
```diff
@@ -1021,13 +1021,34 @@ def self.Pathname(arg)
...
+
+      # Returns the most useful type error possible
+      #
+      # If the object does not respond_to?(:inspect), we return the class, else we
+      # return nil. In all cases, this method is tightly bound to callers, as this
+      # method appends the required space to make the error message look good.
+      #
+      # Used internally by Lotus::Utils::Kernel. This method is not part of the public
+      # API.
+      def self.inspect_type_error(arg)
```
AlfonsoUceda: IMO this method should be private and can you add `@since x.x.x`, `@api private` ? Thanks

AlfonsoUceda:
:+1: errors more useful

lib/lotus/utils/kernel.rb
```diff
@@ -1021,13 +1021,40 @@ def self.Pathname(arg)
...
+        # Returns the most useful type error possible
+        #
+        # If the object does not respond_to?(:inspect), we return the class, else we
+        # return nil. In all cases, this method is tightly bound to callers, as this
+        # method appends the required space to make the error message look good.
+        #
+        # Used internally by Lotus::Utils::Kernel. This method is not part of the public
+        # API.
+        #
+        # @since 0.4.2
```
AlfonsoUceda: please, let @jodosha put 0.4.2 better for now x.x.x, and if you put a space between `@since` and `@private` I'll thank you

francois:
@AlfonsoUceda, I cleaned up as you requested, and squashed all three commits as one. The history will be cleaner this way.

AlfonsoUceda:
@francois perfect ;) good job, :+1:

jodosha:
@francois Thanks for this PR :+1: Helpful error messages are great to have because they helps devs while debugging. Merged. It will be released on tomorrow.

PR #80, opened 2015-06-09
Interactor improvements

### LINKED ISSUE #77: Interactor: introduce `expose` to replace automagic instance variable collection
Hello,

Once again, thanks you all for the great work on Lotus!

At some point on Gitter I mention replace the automagical instance variable conversion to mimic `Lotus::Action` `expose` behavior:

```ruby
class SignIn < Struct.new(:params)
  include Lotus::Interactor

  expose :user

  def call
    @user = User.find params[:id]
  end
end

result = SignIn.new({ id: 15 }).call
result.exposures # => { user: <User:0x007f965c1d0318 @id=15> }
result.user # => <User:0x007f965c1d0318 @id=15>
```

The only difference will be the direct mapping as methods between exposures (as part of payload).

I've a patch for this locally, but I'm concern about the change in behavior, which is why I'm submitting this question first:

Should I keep current behavior (expose all instance variables) *unless* `expose` is used? That way existing code will not break.

What would be the proper approach to handle such scenario?

Looking forward your feedback.

Thank you.

jodosha:
This is a proposal for a few `Lotus::Interactor` improvements.

:warning: **The following change isn't backward compatible** :warning: 

--

### Remove magic :sparkles: in favor of explicitness

Until now all the instance variables from an interactor were automatically exposed into the returning value of `#call`, which is an instance of `Lotus::Interactor::Result`.

#### Before

```ruby
class Signup
  include Lotus::Interactor

  def initialize(params)
    @params = params
    @user = User.new(params.fetch(:user))
  end

  def call
    # ...
  end
end

result = Signup.new(user: { name: "Luca" }).call

result.user   # => #<User:0x007fa311105778 @id=1 @name="Luca">
result.params # => { :user=>{:name=>"Luca"} }
```

#### After

```ruby
class Signup
  include Lotus::Interactor
  expose :user

  def initialize(params)
    @params = params
    @user = User.new(params.fetch(:user))
  end
end

result = Signup.new(user: { name: "Luca" }).call

result.user   # => #<User:0x007fa311105778 @id=1 @name="Luca">
result.params # => raises NoMethodError because we haven't exposed it
```

---

/cc @lotus/core-team @AlfonsoUceda @luislavena 

Closes #77

### DIFF
--- a/lib/lotus/interactor.rb
+++ b/lib/lotus/interactor.rb
@@ -68,6 +68,7 @@ module Lotus
       def add_error(*errors)
         @errors << errors
         @errors.flatten!
+        nil
       end
 
       # Returns the first errors collected during an operation
@@ -156,8 +157,8 @@ module Lotus
       #   class UpdateProfile
       #     include Lotus::Interactor
       #
-      #     def initialize(person, params)
-      #       @person = person
+      #     def initialize(user, params)
+      #       @user   = user
       #       @params = params
       #     end
       #
@@ -181,11 +182,12 @@ module Lotus
       #
       # @raise [NoMethodError] if this isn't implemented by the including class.
       #
-      # @example Instance variables in result payload
+      # @example Expose instance variables in result payload
       #   require 'lotus/interactor'
       #
       #   class Signup
       #     include Lotus::Interactor
+      #     expose :user, :params
       #
       #     def initialize(params)
       #       @params = params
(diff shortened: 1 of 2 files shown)


luislavena:
@jodosha the changes looks much better than the ones I had :smile: 

The only difference is that I considered backward compatibility to keep `_instance_variables` in case `_exposures` were empty.

If is ok to break that compatibility then :heart: :smile: 

Once again thank you for taking the time and sorry not being able to deliver as promised :cry:

jodosha:
@luislavena We're gonna bump v0.5.0 for lotus-utils and write about these breaking changes in the lotusrb v0.4.0 ANN.

PR #82, opened 2015-06-30
Bug in PathPrefix's relative_join

theocodes:
This is an issue i tracked back to from the router.

So given this prefix:
```
prefix = Lotus::Utils::PathPrefix.new('fruits')
```

when joined with the same string

```
prefix.join('fruits') # => /fruits
```
will get rid of one of them only keeping the above `/fruits` instead of `/fruits/fruits`

Obviously that shouldn't be the case because we should be able to have paths such as `/fruits/fruits/fruits/fruits` if we wanted to.

Basically the issue is because PathPrefix uses Lotus::Kernel.Array which calls .uniq! on itself and removes any duplicate items.

I was able to fix the issue by using the standard array for the relative_join + calling chomp("/") on the result of the join to make some tests happy. 

I'm not entirely sold on this solution because it seems to me that even though this fixed the problem entirely as far as I can tell (all tests pass*), Kernel.Array was put there for a reason beyond my understanding at this point. So please let me know what you think!

*While all tests pass on Utils, Router, Lotus and the example app with the initial problem (linked below), I had to amend 2 tests on Router which probably needs review since i'm not very happy with changing tests without knowing exactly what it is that it's testing. 

I've added failing tests and solution to both Utils and Router. Please let me know what you think and if you need me to clarify anything.

Link to the app with the initial routing issue + failing tests:
https://github.com/theocodes/routertest

Thanks!

### DIFF
--- a/lib/lotus/utils/path_prefix.rb
+++ b/lib/lotus/utils/path_prefix.rb
@@ -82,7 +82,7 @@ module Lotus
         prefix = @string.gsub(@separator, separator)
 
         self.class.new(
-          [prefix, strings].join(separator).chomp("/"),
+          [prefix, strings].join(separator).chomp(separator),
           separator
         ).relative!
       end
--- a/lib/lotus/utils/path_prefix.rb
+++ b/lib/lotus/utils/path_prefix.rb
@@ -82,7 +82,7 @@ module Lotus
         prefix = @string.gsub(@separator, separator)
 
         self.class.new(
-          [prefix, strings].join(separator).chomp(separator),
+          [prefix, strings].flatten.compact.join(separator),
           separator
         ).relative!
       end
--- a/lib/lotus/utils/path_prefix.rb
+++ b/lib/lotus/utils/path_prefix.rb
@@ -82,7 +82,7 @@ module Lotus
         prefix = @string.gsub(@separator, separator)
 
         self.class.new(
-          Utils::Kernel.Array([prefix, strings]).join(separator),
+          [prefix, strings].join(separator).chomp("/"),
           separator
         ).relative!
       end
--- a/test/path_prefix_test.rb
+++ b/test/path_prefix_test.rb
@@ -29,6 +29,11 @@ describe Lotus::Utils::PathPrefix do
       prefix.join('/cherries').must_equal '/fruits/cherries'
     end
 
+    it 'joins a string that is the same as the prefix' do
+      prefix = Lotus::Utils::PathPrefix.new('fruits')
+      prefix.join('fruits').must_equal '/fruits/fruits'
+    end
+
     it 'joins a string when the root is blank' do
       prefix = Lotus::Utils::PathPrefix.new
       prefix.join('tea').must_equal '/tea'


lib/lotus/utils/path_prefix.rb
```diff
@@ -82,7 +82,7 @@ def relative_join(strings, separator = @separator)
         prefix = @string.gsub(@separator, separator)
 
         self.class.new(
-          Utils::Kernel.Array([prefix, strings]).join(separator),
+          [prefix, strings].join(separator).chomp("/"),
```
theocodes: It might be that since uniq is the breaking part of Kernel.Array here. I could keep this 'compact' and 'flattened' and just leave the 'uniq' out. Eg.
```
[prefix, strings].flatten.compact.join(separator).chomp("/")
```
What you think?

AlfonsoUceda: @jodosha is there a reason to use `Utils::Kernel.Array` here?

jodosha: @theocodes Can you please replace `"/"` with a constant? Thank you.

jodosha: @AlfonsoUceda I can't recall. Have you guys tried to run all the other test suites against this branch? If all they pass, we can merge this.

AlfonsoUceda: cc @theocodes can you test this?

theocodes: @jodosha @AlfonsoUceda  I have personally ran against router, lotus and (obviously) utils. I will run it against the rest of the gems in a bit to ensure.

theocodes: @jodosha would it make sense to pass the `separator` that is local to `relative_join` to chomp? or actually use `DEFAULT_SEPARATOR`?

Reason I ask is because I'm not familiar with all the use cases of that method, but I wonder if we were handling a path with '#' for separator, if it would make sense to use 'DEFAULT_SEPARATOR` which is '/' and not '#'? 

Hope this makes sense?

lib/lotus/utils/path_prefix.rb
```diff
@@ -82,7 +82,7 @@ def relative_join(strings, separator = @separator)
         prefix = @string.gsub(@separator, separator)
 
         self.class.new(
-          Utils::Kernel.Array([prefix, strings]).join(separator),
+          [prefix, strings].flatten.compact.join(separator),
```
theocodes: This now follows the same pattern as Kernal.Array (as previously) but without the offending part (uniq).

AlfonsoUceda:
this is ok for me
@jodosha @joneslee85

lib/lotus/utils/path_prefix.rb
theocodes: I ran all this against all tests suites and all is green.

jodosha:
@theocodes @AlfonsoUceda Thank you :+1: Merging this.

PR #86, opened 2015-09-21
Add lib/lotus-utils.rb

joneslee85:


jodosha:
@joneslee85 Is this for Bundler/Rubygems compat? :+1:

joneslee85:
yup

PR #88, opened 2015-10-02
Fix Encoding::UndefinedConversionError in encode method.

constXife:
I defined following code in controller.
```ruby
module Web::Controllers::Home
  class Index
    # ...
    def call(params)
      @tag = params[:tag]
    end
  end
end
```
Then I received Encoding::UndefinedConversionError in template from method self.encode in Lotus::Utils when tag contain cyrillic symbols. For some reason params[:tag] have ASCII-8BIT encoding.

So, I added quick fix for that issue.

### DIFF
--- a/lib/lotus/utils/escape.rb
+++ b/lib/lotus/utils/escape.rb
@@ -529,6 +529,8 @@ module Lotus
       # @api private
       def self.encode(input)
         input.to_s.encode(Encoding::UTF_8)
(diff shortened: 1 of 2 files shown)


joneslee85:
cc @jodosha

jodosha:
Hello @constXife Thanks for your precious fix. Merged :+1:

PR #94, opened 2015-11-30
Introducing Lotus::Utils::String#rsub

jodosha:
It replaces rightmost occurence of given pattern and replaces with given replacement.

```ruby
require 'lotus/utils/string'

string = Lotus::Utils::String.new('authors/books/index')
string.rsub("/", "#") # => "authors/books#index"
```

This feature supports https://github.com/lotus/router/pull/76

/cc @lotus/core-team

### DIFF
--- a/lib/lotus/utils/string.rb
+++ b/lib/lotus/utils/string.rb
@@ -340,6 +340,40 @@ module Lotus
         end
       end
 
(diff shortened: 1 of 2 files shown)


AlfonsoUceda:
you rocks ;) nice :+1: but check CI please

PR #95, opened 2015-12-15
Reduce allocations for String#classify method

davydovanton:
I used stackprof gem and this code:
``` ruby
  100_000.times { Lotus::Utils::String.new('lotus_utils/test').classify }
```

### Before
```
==================================
  Mode: object(1)
  Samples: 3500005 (0.00% miss rate)
  GC: 0 (0.00%)
==================================
     TOTAL    (pct)     SAMPLES    (pct)     FRAME
   3300005  (94.3%)     1600004  (45.7%)     Lotus::Utils::String#classify
    900000  (25.7%)      900000  (25.7%)     Lotus::Utils::String#method_missing
    800001  (22.9%)      800001  (22.9%)     Lotus::Utils::String#split
```

### After
```
==================================
  Mode: object(1)
  Samples: 3000005 (0.00% miss rate)
  GC: 0 (0.00%)
==================================
     TOTAL    (pct)     SAMPLES    (pct)     FRAME
   2800005  (93.3%)     1100004  (36.7%)     Lotus::Utils::String#classify
    900000  (30.0%)      900000  (30.0%)     Lotus::Utils::String#scan
    800001  (26.7%)      800001  (26.7%)     Lotus::Utils::String#split
```

:tada:

### DIFF
--- a/lib/lotus/utils/string.rb
+++ b/lib/lotus/utils/string.rb
@@ -6,6 +6,12 @@ module Lotus
     #
     # @since 0.1.0
     class String
+      # Empty string for #classify
+      #
+      # @since 0.6.0
+      # @api private
+      EMPTY_STRING        = ''.freeze
+
       # Separator between Ruby namespaces
       #
       # @since 0.1.0
@@ -135,14 +141,14 @@ module Lotus
       #   string = Lotus::Utils::String.new 'lotus_utils'
       #   string.classify # => 'LotusUtils'
       def classify
-        words = split(CLASSIFY_WORD_SEPARATOR).map(&:capitalize)
+        words = split(CLASSIFY_WORD_SEPARATOR).map!(&:capitalize)
         delimiters = scan(CLASSIFY_WORD_SEPARATOR)
 
         delimiters.map! do |delimiter|
-          delimiter == CLASSIFY_SEPARATOR ? nil : NAMESPACE_SEPARATOR
+          delimiter == CLASSIFY_SEPARATOR ? EMPTY_STRING : NAMESPACE_SEPARATOR
         end
 
-        self.class.new words.zip(delimiters).compact.join
+        self.class.new words.zip(delimiters).join
       end
 
       # Return a downcased and underscore separated version of the string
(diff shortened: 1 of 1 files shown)


lib/lotus/utils/string.rb
```diff
@@ -135,14 +141,14 @@ def capitalize
       #   string = Lotus::Utils::String.new 'lotus_utils'
       #   string.classify # => 'LotusUtils'
       def classify
-        words = split(CLASSIFY_WORD_SEPARATOR).map(&:capitalize)
+        words = split(CLASSIFY_WORD_SEPARATOR).map!(&:capitalize)
         delimiters = scan(CLASSIFY_WORD_SEPARATOR)
 
         delimiters.map! do |delimiter|
-          delimiter == CLASSIFY_SEPARATOR ? nil : NAMESPACE_SEPARATOR
+          delimiter == CLASSIFY_SEPARATOR ? EMPTY_STRING : NAMESPACE_SEPARATOR
```
marshall-lee: You introduced `EMPTY_STRING` just to remove a `compact` call?
```ruby
['s',nil,'urp',nil,'rise'].join
# => "surprise"
```

lib/lotus/utils/string.rb
```diff
@@ -135,14 +141,14 @@ def capitalize
       #   string = Lotus::Utils::String.new 'lotus_utils'
       #   string.classify # => 'LotusUtils'
       def classify
-        words = split(CLASSIFY_WORD_SEPARATOR).map(&:capitalize)
+        words = split(CLASSIFY_WORD_SEPARATOR).map!(&:capitalize)
```
joneslee85: personally I am not a big fan of mutating data

marshall-lee: @joneslee85 but it's merely temporary data.

davydovanton: yep, I agree with @marshall-lee, `split` returns new data object on each call, thats why I prefer `map!` here.

lib/lotus/utils/string.rb
davydovanton: yes, but empty string is more idiomatically here I think

lib/lotus/utils/string.rb
jodosha: yes, that mutation isn't relevant, but it helps to speed up. :+1:

jodosha:
@davydovanton Thank you very much, this is the kind of contributions that I really :heart:

davydovanton:
thanks for merge, @jodosha :star2:

PR #102, opened 2016-01-01
Fixed jruby head tests related with bigdecimal

AlfonsoUceda:
Removed jruby-9000 from travis and use only jruby-head.

Fixed tests related with BigDecimal and jruby-head

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ script: 'bundle exec rake test:coverage --trace'
 rvm:
   - 2.2.4
   - 2.3.0
-  - jruby-9000
+  - jruby-9.0.5.0
   - ruby-head
   - jruby-head
 
@@ -13,4 +13,3 @@ matrix:
   allow_failures:
     - rvm: ruby-head
     - rvm: jruby-head
-    - rvm: jruby-9000 # See https://github.com/department-of-veterans-affairs/connect_vbms/issues/15 and https://github.com/rvm/rvm/issues/3475
--- a/lib/hanami/utils/kernel.rb
+++ b/lib/hanami/utils/kernel.rb
@@ -25,6 +25,14 @@ module Hanami
       # @see Hanami::Utils::Kernel.Integer
       NUMERIC_MATCHER = /\A([\d\/\.\+iE]+|NaN|Infinity)\z/.freeze
 
+      # Default precision for bigdecimal with rational number
+      #
+      # @since x.x.x
+      # @api private
+      #
+      # @see Hanami::Utils::Kernel.BigDecimal
+      BIGDECIMAL_DEFAULT_PRECISION = 1
+
       # Coerces the argument to be an Array.
       #
       # It's similar to Ruby's Kernel.Array, but it applies further
@@ -334,6 +342,7 @@ module Hanami
       # Coerces the argument to be a BigDecimal.
       #
       # @param arg [Object] the argument
+      # @param precision [Keyword] precision for Rational objects (Only JRuby).
       #
       # @return [BigDecimal] the result of the coercion
       #
@@ -404,15 +413,14 @@ module Hanami
       #   # Missing #respond_to?
       #   input = BasicObject.new
       #   Hanami::Utils::Kernel.BigDecimal(input) # => TypeError
-      def self.BigDecimal(arg)
+      def self.BigDecimal(arg, precision: BIGDECIMAL_DEFAULT_PRECISION)
         case arg
+        when ->(a) { a.respond_to?(:to_d) && a.class == Rational } then arg.to_d(precision)
         when ->(a) { a.respond_to?(:to_d) } then arg.to_d
         when Float, Complex, Rational
           BigDecimal(arg.to_s)
-        when ->(a) { a.to_s.match(NUMERIC_MATCHER) }
-          BigDecimal.new(arg)
         else
-          raise TypeError.new "can't convert #{inspect_type_error(arg)}into BigDecimal"
+          BigDecimal.new(arg)
         end
       rescue NoMethodError
         raise TypeError.new "can't convert #{inspect_type_error(arg)}into BigDecimal"
--- a/test/kernel_test.rb
+++ b/test/kernel_test.rb
@@ -812,10 +812,16 @@ describe Hanami::Utils::Kernel do
       end
 
       describe 'when a rational number is given' do
-        let(:input) { Rational(0.3) }
+        let(:input)     { Rational(0.3) }
+        let(:precision) { 1 }
 
         it 'returns an BigDecimal' do
-          @result.must_equal BigDecimal.new(input.to_s)
+          if Hanami::Utils.jruby?
+            @result = Hanami::Utils::Kernel.BigDecimal(input, precision: precision)
+            @result.must_equal BigDecimal.new(input, precision)
+          else
+            @result.must_equal BigDecimal.new(input.to_s)
+          end
         end
       end
 
@@ -859,6 +865,22 @@ describe Hanami::Utils::Kernel do
         it 'returns a BigDecimal' do
           @result.must_equal BigDecimal.new(23)
         end
+      end
+    end
+
+    describe 'when a string without numbers is given' do
+      let(:input) { 'home' }
+
+      it 'returns an BigDecimal' do
+        Hanami::Utils::Kernel.BigDecimal(input).must_equal BigDecimal.new(0)
+      end
+    end
+
+    describe 'when a string which starts with a big decimal is given' do
+      let(:input) { '23.0 street' }
+
+      it 'returns an BigDecimal' do
+        Hanami::Utils::Kernel.BigDecimal(input).must_equal BigDecimal.new(23)
       end
     end
 
@@ -908,24 +930,6 @@ describe Hanami::Utils::Kernel do
 
         it 'raises error' do
           -> { Hanami::Utils::Kernel.BigDecimal(input) }.must_raise(TypeError)
-        end
-      end
-
-      describe 'when a string without numbers is given' do
-        let(:input) { 'home' }
-
-        it 'raises error' do
-          exception = -> { Hanami::Utils::Kernel.BigDecimal(input) }.must_raise(TypeError)
-          exception.message.must_equal "can't convert #{input.inspect} into BigDecimal"
-        end
-      end
-
-      describe 'when a string which starts with a big decimal is given' do
-        let(:input) { '23.0 street' }
-
-        it 'raises error' do
-          exception = -> { Hanami::Utils::Kernel.BigDecimal(input) }.must_raise(TypeError)
-          exception.message.must_equal "can't convert #{input.inspect} into BigDecimal"
         end
       end
 
--- a/test/escape_test.rb
+++ b/test/escape_test.rb
@@ -350,8 +350,9 @@ describe Hanami::Utils::Escape do
             end
 
             it "escapes a Time" do
+              time_string = Hanami::Utils.jruby? ? "2016-01-27 12:00:00 UTC" : "2016-01-27 12:00:00 +0000"
               result = mod.html(Time.new(2016,01,27, 12, 0, 0, 0))
-              result.must_equal "2016-01-27 12:00:00 +0000"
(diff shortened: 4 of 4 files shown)


jodosha:
@AlfonsoUceda Can you please link here the jruby bug? and explain the reason of this? As future memory. TY!

.travis.yml
```diff
@@ -19,7 +19,6 @@ rvm:
   - 2.2.3
   - 2.2.4
   - 2.3.0
-  - jruby-9000
```
jodosha: @AlfonsoUceda Why this was removed?

test/kernel_test.rb
```diff
@@ -816,10 +816,10 @@ def to_d
       end
 
       describe 'when a rational number is given' do
-        let(:input) { Rational(0.3) }
+        let(:input) { Rational(0.3).to_s }
```
jodosha: @AlfonsoUceda Transforming the input in a string, changes the purpose of this test. See the line above `"when a rational number is given"`.

AlfonsoUceda: you are right, I have to fix it

.travis.yml
AlfonsoUceda: I removed because jruby-9000 is broken due to BigDecimal, but I have to add the right version

lib/lotus/utils/kernel.rb
```diff
@@ -418,8 +426,9 @@ def self.Integer(arg)
       #   # Missing #respond_to?
       #   input = BasicObject.new
       #   Lotus::Utils::Kernel.BigDecimal(input) # => TypeError
-      def self.BigDecimal(arg)
+      def self.BigDecimal(arg, precision: DEFAULT_PRECISION)
         case arg
+        when ->(a) { a.respond_to?(:to_d) && a.class == Rational } then arg.to_d(precision)
```
AlfonsoUceda: @jodosha take a look at this line. We need add precision for BigDecimal when you are using Rational number, because only a Rational number needs an argument for `#to_d` method I added this line, what do you think? I don't like so much this solution.

AlfonsoUceda:
@jodosha we can support lotus/utils from jruby 9.0.5.0 (jruby-head) because in previous versions complex are broken if we pass a complex to a big decimal.

lib/lotus/utils/kernel.rb
jodosha: @AlfonsoUceda Hey thanks for looking at this.

1. `DEFAULT_PRECISION` it isn't a descriptive name. We know that is about BigDecimal, but only because we're discussing in this PR. By looking at that, there is no mention of this. What if we need to introduce a precision for Float? Shall we specify `BIGDECIMAL_DEFAULT_PRECISION` instead?
2. Applying `precision:` only to Rational is misleading. From an outside perspective would expect that precision is always applied, regardless of the input. Can we apply for all the kind of inputs? Which precision should we choose by default?

joneslee85:
@AlfonsoUceda I am with you on this

jodosha:
For reference, I've checked which jruby version introduced this change and it was 9.0.1.0. I'm :+1: to bump to 9.0.5.0 as minimum supported version.

test/kernel_test.rb
```diff
@@ -918,18 +924,30 @@ def to_d
       describe 'when a string without numbers is given' do
         let(:input) { 'home' }
 
-        it 'raises error' do
-          exception = -> { Lotus::Utils::Kernel.BigDecimal(input) }.must_raise(TypeError)
-          exception.message.must_equal "can't convert #{input.inspect} into BigDecimal"
+        unless Lotus::Utils.jruby?
+          it 'raises error' do
```
jodosha: This test is bypassed because jruby doesn't raise any error in this case, but it should because MRI does that.

AlfonsoUceda: but... we have to deal for now like this, maybe in the future we raise an error or do we should report this?

lib/lotus/utils/kernel.rb
AlfonsoUceda: @jodosha for JRuby only Rational has `#to_d` method with one parameter, that parameter is the precision.

test/kernel_test.rb
joneslee85: IMO we should report it

lib/hanami/utils/kernel.rb
```diff
@@ -404,15 +413,14 @@ def self.Integer(arg)
       #   # Missing #respond_to?
       #   input = BasicObject.new
       #   Hanami::Utils::Kernel.BigDecimal(input) # => TypeError
-      def self.BigDecimal(arg)
+      def self.BigDecimal(arg, precision: BIGDECIMAL_DEFAULT_PRECISION)
```
pascalbetz: FYI: a ```case``` with block is considerably slower than ```if/elsif```.

jodosha: @pascalbetz Do you have a benchmark for this?

lib/hanami/utils/kernel.rb
```diff
@@ -25,6 +25,14 @@ module Kernel
       # @see Hanami::Utils::Kernel.Integer
       NUMERIC_MATCHER = /\A([\d\/\.\+iE]+|NaN|Infinity)\z/.freeze
 
+      # Default precision for bigdecimal with rational number
+      #
+      # @since x.x.x
+      # @api private
+      #
+      # @see Hanami::Utils::Kernel.BigDecimal
+      BIGDECIMAL_DEFAULT_PRECISION = 1
```
jodosha: @AlfonsoUceda The only thing I'm not sure about is the value of this constant. By using this low precision, we drop accuracy, and accuracy is what you look for when using BigDecimal. Shall we consider a bigger value for this? Eg. `16`.

AlfonsoUceda: you are right, when you use big decimal is for big precision. I don't know if a good value is 16 but I think is big accuracy  hehe, I'll change with that value

lib/hanami/utils/kernel.rb
```diff
@@ -404,15 +413,14 @@ def self.Integer(arg)
       #   # Missing #respond_to?
       #   input = BasicObject.new
       #   Hanami::Utils::Kernel.BigDecimal(input) # => TypeError
-      def self.BigDecimal(arg)
+      def self.BigDecimal(arg, precision: BIGDECIMAL_DEFAULT_PRECISION)
         case arg
+        when ->(a) { a.respond_to?(:to_d) && a.class == Rational } then arg.to_d(precision)
```
jodosha: @AlfonsoUceda I discovered that on MRI `Rational` doesn't respond to `#to_d` and this double check can be perf consuming. What if we flip the case conditionals like this?

```ruby
def self.BigDecimal(arg)
        case arg
        when Float, Complex, Rational
          BigDecimal(arg.to_s)
        when ->(a) { a.respond_to?(:to_d) }
           arg.to_d
        else
          BigDecimal.new(arg)
        end
      rescue NoMethodError
        raise TypeError.new "can't convert #{inspect_type_error(arg)} into BigDecimal"
      end
end
```

This should make the test to pass, but avoids the need for us to decide the precision with that constant. What do you think?

AlfonsoUceda: @jodosha in JRuby Rational needs an argument, that argument is the precision, the other object responds to_d method but the don't need the precision.

jodosha: @AlfonsoUceda Sure, but because `Rational` in case statement appears before of the check for `#to_d`, it will never be hit, unless `arg` is an object that isn't a `Float`, or a `Complex` or a `Rational`.

jodosha: @AlfonsoUceda I'd replace `when Float, Complex, Rational` with `when Numeric`

lib/hanami/utils/kernel.rb
pascalbetz: @jodosha https://gist.github.com/pascalbetz/403a60ae6dd9a3089b1e

lib/hanami/utils/kernel.rb
pascalbetz: @AlfonsoUceda so what you are interested in is the arity, not the class the method is defined on?

jodosha: @AlfonsoUceda @pascalbetz This arity problem will disappear if we apply this patch:

```diff
diff --git a/lib/hanami/utils/kernel.rb b/lib/hanami/utils/kernel.rb
index eecd412..d49e2f6 100644
--- a/lib/hanami/utils/kernel.rb
+++ b/lib/hanami/utils/kernel.rb
@@ -25,14 +25,6 @@ module Hanami
       # @see Hanami::Utils::Kernel.Integer
       NUMERIC_MATCHER = /\A([\d\/\.\+iE]+|NaN|Infinity)\z/.freeze
 
-      # Default precision for bigdecimal with rational number
-      #
-      # @since x.x.x
-      # @api private
-      #
-      # @see Hanami::Utils::Kernel.BigDecimal
-      BIGDECIMAL_DEFAULT_PRECISION = 1
-
       # Coerces the argument to be an Array.
       #
       # It's similar to Ruby's Kernel.Array, but it applies further
@@ -413,12 +405,12 @@ module Hanami
       #   # Missing #respond_to?
       #   input = BasicObject.new
       #   Hanami::Utils::Kernel.BigDecimal(input) # => TypeError
-      def self.BigDecimal(arg, precision: BIGDECIMAL_DEFAULT_PRECISION)
+      def self.BigDecimal(arg)
         case arg
-        when ->(a) { a.respond_to?(:to_d) && a.class == Rational } then arg.to_d(precision)
-        when ->(a) { a.respond_to?(:to_d) } then arg.to_d
-        when Float, Complex, Rational
+        when Numeric
           BigDecimal(arg.to_s)
+        when ->(a) { a.respond_to?(:to_d) }
+          arg.to_d
         else
           BigDecimal.new(arg)
         end
diff --git a/test/kernel_test.rb b/test/kernel_test.rb
index e19eb9d..4c60d60 100644
--- a/test/kernel_test.rb
+++ b/test/kernel_test.rb
@@ -813,15 +813,9 @@ describe Hanami::Utils::Kernel do
 
       describe 'when a rational number is given' do
         let(:input)     { Rational(0.3) }
-        let(:precision) { 1 }
 
         it 'returns an BigDecimal' do
-          if Hanami::Utils.jruby?
-            @result = Hanami::Utils::Kernel.BigDecimal(input, precision: precision)
-            @result.must_equal BigDecimal.new(input, precision)
-          else
-            @result.must_equal BigDecimal.new(input.to_s)
-          end
+          @result.must_equal BigDecimal.new(input.to_s)
         end
       end
 

```

AlfonsoUceda:
@jodosha thanks! sorry I saw the message but I forgot apply it.

lib/hanami/utils/kernel.rb
jodosha: @pascalbetz Interesting, thank you. What about to measure the new implementation at https://github.com/hanami/utils/commit/9fb7bcf036e64d7180539c2e2b4b6e2d37e739cb and eventually refactor with `if/else`? Thank you very much :+1:

jodosha:
@AlfonsoUceda No worries, man :+1:

lib/hanami/utils/kernel.rb
pascalbetz: @jodosha here are the three versions compared (currently released, master, replacing case/when with if/else): https://gist.github.com/pascalbetz/05972b4aab6d5a948c8a

I copied them into one "self-containing" file in order to avoid version troubles with installed gems.

Tested with 2.2.0, with 2.2.2 and jruby. Converting to if/else does not gain us much on jruby, it seems.

pascalbetz: Running my general "if/else" vs. "case/when" benchmark "if/else" is still two times faster, even with Jruby. But of course this boils down to number of Procs are being called..

jodosha: @pascalbetz I've replied to that Gist. The difference between switch and if/else is about 10%. You can use `x.compare!` to let `benchmark-ips` to calculate the diff for you :wink: 

If you have time, feel free to open a PR with this speed improvement.

PR #107, opened 2016-01-15
Create Option class for working with options

davydovanton:
I found [this TODO](https://github.com/lotus/model/blob/master/lib/lotus/model/configuration.rb#L265-L268) and I decided to do it.

### DIFF
--- a/lib/lotus/utils/options.rb
+++ b/lib/lotus/utils/options.rb
@@ -1,0 +1,33 @@
+require 'lotus/utils/hash'
+
+module Lotus
+  module Utils
+    # This class provides methods for options.
+    #
+    # @since x.x.x
+    class Options < Lotus::Utils::Hash
+      # Check options on included key or keys
+      #
+      # @param keys [Symbol] the specific key or keys.
+      #
+      # @return [NilClass] if options contain a keys.
+      #
+      # @raise [ArgumentError] if options not contain a keys.
+      #
+      # @since x.x.x
+      #
+      # @example
+      #   require 'lotus/utils/class'
+      #
+      #   options = Lotus::Utils::Options.new(foo: :bar)
+      #
+      #   options.check_keys(:foo) # => nil
+      #   options.check_keys(:foo, :baz) # => raises ArgumentError
+      def check_options!(*keys)
+        keys.each do |key|
+          raise ArgumentError.new("missing keyword: #{key}") unless self.keys.include?(key)
+        end
+      end
+    end
+  end
+end
--- a/test/options_test.rb
+++ b/test/options_test.rb
@@ -1,0 +1,26 @@
+require 'test_helper'
+require 'lotus/utils/options'
+
+describe Lotus::Utils::Options do
+  let(:options) { Lotus::Utils::Options.new(foo: :bar) }
+
+  describe '#check_options!' do
+    describe 'when options contain keys' do
+      it 'does not raise exeption' do
+        -> {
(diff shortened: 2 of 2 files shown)


lib/lotus/utils/options.rb
```diff
@@ -0,0 +1,33 @@
+require 'lotus/utils/hash'
+
+module Lotus
+  module Utils
+    # This class provide methods for options.
+    #
+    # @since x.x.x
+    class Options < Lotus::Utils::Hash
+      # Check options on inclided key or keys
```
joneslee85: s/inclided/included

lib/lotus/utils/options.rb
```diff
@@ -0,0 +1,33 @@
+require 'lotus/utils/hash'
+
+module Lotus
+  module Utils
+    # This class provide methods for options.
```
joneslee85: s/provide/provides

davydovanton:
@joneslee85 thanks for catch! fixed.

test/options_test.rb
```diff
@@ -0,0 +1,24 @@
...
+          options.check_options!(:foo)
+        }.must_be_silent
+      end
+    end
+
+    describe 'when options not contain a keys' do
+      it 'raises exeption' do
+        -> {
+          options.check_options!(:baz, :foobar)
+        }.must_raise(ArgumentError)
```
joneslee85: can you please also check the exception message?

test/options_test.rb
```diff
@@ -0,0 +1,24 @@
+require 'test_helper'
+require 'lotus/utils/options'
+
+describe Lotus::Utils::Options do
+  let(:options) { Lotus::Utils::Options.new(foo: :bar) }
+
+  describe '#check_options!' do
+    describe 'when options contain a keys' do
```
joneslee85: keys is a plural, I think you should remove 'a' or the 's'

test/options_test.rb
```diff
@@ -0,0 +1,24 @@
...
+      it 'not raises exeption' do
```
joneslee85: s/not/does not

joneslee85: s/raises/raise

davydovanton:
@joneslee85 I added test for error message. Fixed grammar :tada:

joneslee85:
@davydovanton thanks for the PR. Ruby 2.0 will be EOL on Feb 16th 2016, that also means Lotus will be able to use `kwargs`. Taking into the fact that it is pretty close now, I am questioning if we still need this util method.

joneslee85:
cc @jodosha

davydovanton:
@joneslee85 interesting question. So I see two reasons why these changes are helpful:
1. These changes help to avoid duplication of code now
2. We can change this code to `kwargs` method faster in future

jodosha:
@davydovanton Yes, @joneslee85 is right. For that specific use case linked by you, it isn't needed anymore. Can you please check the other similar cases? Thank you!

davydovanton:
so, yes I think this class will be helpful only for old ruby version without kwargs.

jodosha:
@davydovanton Okay, thanks for this PR in the first place and thanks for checking the other use cases. :smile: Soon we'll support only Ruby versions that implements kwargs. Fro this reason I'm closing this. Thank you again :dog:

davydovanton:
No problem :)
I happy to help :)

jodosha:
@davydovanton :)

PR #108, opened 2016-01-18
Create simple template class. Fix DRY in view and mailer modules

davydovanton:
in view:
https://github.com/lotus/view/blob/master/lib/lotus/view/template.rb

in mailer:
https://github.com/lotus/mailer/blob/master/lib/lotus/mailer/template.rb

### DIFF
--- a/lib/lotus/utils/template.rb
+++ b/lib/lotus/utils/template.rb
@@ -1,0 +1,53 @@
+require 'tilt'
+
+module Lotus
+  module Utils
+    # A logic-less template.
+    #
+    # @api private
+    # @since 0.1.0
+    class Template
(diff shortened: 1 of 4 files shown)


jodosha:
@davydovanton Thanks for this PR, but `lotus-utils` shouldn't have any runtime dependency, unless it's for a feature that we want to offer. This proposal just introduces a way to DRY code across frameworks, which doesn't justify the add of this gem.

This framework can be used as inflector, or for service objects from other Ruby apps, which don't need to install `tilt` for pluralize some strings.

Closing this :wink:

davydovanton:
no problem, thanks for review! :blush:

PR #111, opened 2016-01-23
do not pass the block to `String#chars`

y-yagi:
Since Ruby 2.0.0, `String#chars` returns an array instead of an enumerator. Passing a block is deprecated.

ref: https://github.com/ruby/ruby/commit/3f9b0936aa846bdf6984019ca40bc629fe05d929

### DIFF
--- a/lib/hanami/utils/escape.rb
+++ b/lib/hanami/utils/escape.rb
@@ -423,7 +423,7 @@ module Hanami
 
         result = SafeString.new
 
-        input.chars do |chr|
+        input.each_char do |chr|
           result << HTML_CHARS.fetch(chr, chr)
         end
 
@@ -456,7 +456,7 @@ module Hanami
 
         result = SafeString.new
 
-        input.chars do |chr|
+        input.each_char do |chr|
           result << encode_char(chr, HTML_ATTRIBUTE_SAFE_CHARS)
         end
 


joneslee85:
@y-yagi LGTM :+1:

jodosha:
@y-yagi Thanks for this PR, I wasn't aware of this deprecation :smile: 

Quoting [Ruby Docs](http://ruby-doc.org/core-2.2.0/String.html#method-i-chars) for `String#chars`:

> Returns an array of characters in `str`. This is a shorthand for `str.each_char.to_a`.

> If a block is given, which is a deprecated form, works the same as `each_char`.

---

It's inefficient to create an array by iterating the chars and then iterating them again. We should use `#each_char` instead of `chars.each`.

Please have a look at this [benchmark](https://gist.github.com/jodosha/42fae9b05c1a819323f3):

```
Calculating -------------------------------------
          chars.each     4.238k i/100ms
           each_char     5.022k i/100ms
-------------------------------------------------
          chars.each     48.040k (±46.1%) i/s -    194.948k
           each_char     62.546k (±42.9%) i/s -    210.924k in   5.262923s

Comparison:
           each_char:    62546.1 i/s
          chars.each:    48039.9 i/s - 1.30x slower
```

---

Would you please update the PR, by using `#each_char` instead? Thank you :+1:

y-yagi:
@jodosha Thanks for the review and detailed description!

`#each_char` looks good. I updated to use `#each_char`. Thanks.

jodosha:
@y-yagi Thank you very much! :+1:

PR #112, opened 2016-01-26
Allow logger to accept a device

davydovanton:
This is first step for all logger issues like https://github.com/hanami/hanami/issues/481.

/cc @jodosha

### DIFF
--- a/lib/hanami/logger.rb
+++ b/lib/hanami/logger.rb
@@ -5,7 +5,7 @@ module Hanami
   # Hanami logger
   #
   # Implement with the same interface of Ruby std lib `Logger`.
-  # It uses `STDOUT` as output device.
+  # It uses `STDOUT`, `STDERR`, file name or open file as output device.
   #
   #
   #
@@ -67,6 +67,13 @@ module Hanami
   #
   #   Hanami::Logger.new('FOO').info('Hello')
   #   # => I, [2015-01-10T21:55:12.727259 #80487]  INFO -- [FOO] : Hello
+  #
+  # @example Write to file
+  #   require 'hanami'
+  #
+  #   Hanami::Logger.new(log_device: 'logfile.log').info('Hello')
+  #   # in logfile.log
+  #   # => I, [2015-01-10T21:55:12.727259 #80487]  INFO -- [FOO] : Hello
   class Logger < ::Logger
     # Hanami::Logger default formatter
     #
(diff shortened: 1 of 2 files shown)


davydovanton:
@hanami/core-team could you check this PR? I can't complete all logger issues without this :disappointed:

jodosha:
@davydovanton Hey, thanks for this PR. :smile: 

There are a couple of improvements to the tests and a more use cases to cover.

Tests:

  * The log file shouldn't be added to the source code.
  * We should use a `tmp/` directory.
  * Ensure that the file is removed in a `before` block
  * When testing an already existing file we can use `FileUtils.touch` to create it under `tmp/`
  * We should assert that the UNIX permissions are the same that `Logger` (from stdlib) uses.

Features:

  * Accept a relative path
  * Accept `Pathname`
  * Accept `IO`/`StringIO`
  * Make `#close` safe when the internal device is `STDOUT` (see below).

---

### STDOUT close

```ruby
require 'logger'

logger = Logger.new(STDOUT)
logger.close

puts "hi"

__END__

➜ ruby logging.rb
logging.rb:6:in `write': closed stream (IOError)
	from logging.rb:6:in `puts'
	from logging.rb:6:in `puts'
	from logging.rb:6:in `<main>'
```

davydovanton:
@jodosha thanks for review. I have some questions:
1. What do you expect with `Pathname`?
2. I check `IO` in `File.new` test. Do you want to check pure `IO` class too?

Thanks!

jodosha:
@davydovanton Thanks for the progress :+1: 

> 1. What do you expect with Pathname?

Same values that we expect from a path represented as a string: relative and absolute path.

> 2. I check IO in File.new test. Do you want to check pure IO class too?

Yes please because `File` is backed by filesystem and it was mandatory to test. While `IO` or `StringIO` are really nice to have because useful for testing purposes or for extensibility with other third party gems.

davydovanton:
@jodosha done. I added tests for `IO` and `Pathname` class too.

lib/hanami/logger.rb
```diff
@@ -122,6 +133,13 @@ def application_name
       @application_name || _application_name_from_namespace || _default_application_name
     end
 
+    # Close the logging device if this device isn't an STDOUT
+    #
+    # @since x.x.x
+    def close
+      super if @log_device != STDOUT
```
jodosha: @davydovanton We should also guard against `$stdout`.

```ruby
super unless [STDOUT, $stdout].include?(@log_device)
```

davydovanton: yep, you're right. Thanks for catch! :+1:

davydovanton:
@jodosha done :+1:

jodosha:
@davydovanton Thank you. Merged :+1:

davydovanton:
@jodosha thanks for merge and review :+1:

PR #116, opened 2016-02-01
Utils::Hash #symbolize! and #stringify! supporst nested Utils::Hash values

thhermansen:
Hi,

This pull request adds support for `symbolize!` and `stringify!`with nested `Hanami::Utils::Hash` as values.

I'm not sure if there is a reason not to do this, or maybe there are a better way of implementing this. For instance: [ActiveSupport::TimeWithZone](https://github.com/rails/rails/blob/0e528b626cb164153dea45b4528888ef663cad83/activesupport/lib/active_support/time_with_zone.rb#L334) do override `is_a?` so it returns `true` when asked if `time_with_zone.is_a? Time`. Some tests broke when went this way with `Utils::Hash`.

### DIFF
--- a/lib/hanami/utils/hash.rb
+++ b/lib/hanami/utils/hash.rb
@@ -65,7 +65,7 @@ module Hanami
       def symbolize!
         keys.each do |k|
           v = delete(k)
-          v = Hash.new(v).symbolize! if v.is_a?(::Hash)
+          v = Hash.new(v).symbolize! if v.is_a?(::Hash) || v.is_a?(::Hanami::Utils::Hash)
 
           self[k.to_sym] = v
         end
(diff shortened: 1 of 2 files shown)


davydovanton:
LGTM :+1:

joneslee85:
:+1:

jodosha:
@thhermansen This looks good, how did you noticed this? By using params?

thhermansen:
Hi!

No, If I remember correctly I think I loaded a JSON column from postgres and all keys were strings. Some values in the JSON was a new JSON object, thus became a new ruby hash and in total a nested Hash in Ruby.

I wanted all keys to be symbols for easier access and tried using `Utils::Hash#symbolize` to accomplish this.

jodosha:
@thhermansen Merged in master. Thank you! :+1:

thhermansen:
No worries, I'm the one who should be thanking you and the Hanami team :)

PR #117, opened 2016-02-01
kernel blank? method

accessd:


joneslee85:
@accessd thanks for the PR. What are the usecases for this new method?

lib/hanami/utils/kernel.rb
```diff
@@ -1022,6 +1022,28 @@ def self.inspect_type_error(arg)
         end
       end
 
+      # Checks object is blank
+      #
+      # @param object the argument
+      #
+      # @return [TrueClass,FalseClass]
```
davydovanton: I think we need to add example of usages. What do you think?

accessd:
@joneslee85 For example, in https://github.com/hanami/hanami/pull/467 I need `blank?` method for `Hanami::Utils::String`. There @jodosha told me that I could use `Lotus::Utils::String.new(value).blank?` method, but I couldn't find it.
So, I thought to add it to `Hanami::Utils::String`, but I noticed that you had suggested add `blank?` to `Hanami::Utils::Kernel` in https://github.com/hanami/utils/pull/66

Also, `blank?` can be useful in `Hanami::Utils::Inflector#pluralize` and `Hanami::Utils::Inflector#singularize`

lib/hanami/utils/kernel.rb
accessd: @davydovanton good idea!

test/kernel_test.rb
```diff
@@ -2508,4 +2510,22 @@ def to_sym
       end
     end
   end
+
+  describe '.blank?' do
+    [nil, false, '', '   ', "  \n\t  \r ", '　', "\u00a0", [], {},
+     Hanami::Utils::Kernel.Boolean(0), Hanami::Utils::String.new(''),
+     Hanami::Utils::Hash.new({})].each do |v|
+       it 'true' do
```
joneslee85: s/true/returns true

test/kernel_test.rb
```diff
@@ -2508,4 +2510,22 @@ def to_sym
...
+         Hanami::Utils::Kernel.blank?(v).must_equal true
+       end
+     end
+
+     [Object.new, true, 0, 1, 'a', [nil], {nil => 0},
+      Hanami::Utils::Kernel.Symbol(:hello), Hanami::Utils::String.new('foo'),
+      Hanami::Utils::Hash.new({foo: :bar})].each do |v|
+        it 'false' do
```
joneslee85: s/false/returns false

joneslee85:
@accessd I'd leave the decision on merging this PR to @jodosha

accessd:
updated

jodosha:
@accessd Thanks for this PR. While I see useful to abstract this concept for all our gems, I don't think `Utils::Kernel` is a good place for it.

I would introduce `Utils::Blank.blank?` for the purpose. Can you please change it and rebase? Thank you. :cat:

accessd:
@jodosha done!

davydovanton:
@jodosha @accessd what do you think about `Utils::Object` in this case? I think this class name is better for general methods like `#blank?`.

jodosha:
@accessd Hey, I've just merged this. Thank you! :+1:

PR #121, opened 2016-02-25
Add level option to logger initializer

davydovanton:
Reference to https://github.com/hanami/hanami/pull/530


davydovanton:
@hanami/core-team

jodosha:
@davydovanton Thanks for the quick implementation :100: I've merged both in master and stable (`0.8.x`). You can now use it with https://github.com/hanami/hanami/pull/530

PR #122, opened 2016-03-13
Introduce Base and JSON log formatters

davydovanton:
Reference to https://github.com/hanami/hanami/issues/483

### DIFF
--- a/lib/hanami/logger.rb
+++ b/lib/hanami/logger.rb
@@ -1,3 +1,4 @@
+require 'json'
 require 'logger'
 require 'hanami/utils/string'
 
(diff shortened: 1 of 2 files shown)


lib/hanami/logger.rb
```diff
@@ -91,8 +92,33 @@ class Formatter < ::Logger::Formatter
       #
       # @see http://www.ruby-doc.org/stdlib/libdoc/logger/rdoc/Logger/Formatter.html#method-i-call
       def call(severity, time, progname, msg)
-        progname = "[#{@application_name}] #{progname}"
-        super(severity, time.utc, progname, msg)
+        hash = {
+          application_name: @application_name,
```
jodosha: @davydovanton What about a shorter `app` instead of `application_name`?

davydovanton: good idea :+1:

lib/hanami/logger.rb
```diff
@@ -91,8 +92,33 @@ class Formatter < ::Logger::Formatter
...
+          severity: severity,
+          time: time.utc,
+          progname: progname
```
jodosha: @davydovanton Is `progname` still used and/or useful? It's always null in the tests.

lib/hanami/logger.rb
```diff
@@ -91,8 +92,33 @@ class Formatter < ::Logger::Formatter
...
+      # @api private
+      def _message_hash(message)
+        case message
+        when Hash
+          message
+        when Exception
+          {
+            message: message.message,
+            backtrace: message.backtrace || [],
+            error_class: message.class
```
jodosha: @davydovanton what about `error`, instead of `error_class`?

lib/hanami/logger.rb
davydovanton: So, in all my tests and projects `progname` was null. I'll remove this

davydovanton:
@jodosha done. I renamed keys and deleted `progname` unnecessary value :tada:

davydovanton:
@jodosha updated documentation examples and added part about formatters 👍

jodosha:
@davydovanton Thank you very much. Just merged! 👍

davydovanton:
amazing 🎉 
Thanks for helping 👍

joneslee85:
Great work. Thanks @davydovaton

PR #124, opened 2016-03-20
Fix singularization of plural words ending with rses

TheSmartnik:
Words like [morses](https://en.wiktionary.org/wiki/morses#English), verses etc. were singularized incorrectly. This should fix the issue


jodosha:
@TheSmartnik Thank you, merging this :+1:

PR #126, opened 2016-03-21
Fix ruby warnings

jodosha:
Fixing a few Ruby warnings due to Rake 11.

---

/cc @hanami/core-team @hanami/contributors


davydovanton:
LGTM :+1:

Issue #128, opened 2016-03-21
URI.unescape is obsolete


jodosha:
`Hanami::Utils::Escape.url` is using `URI.unescape` (alias of `URI.decode`), which is deprecated. We need to update the implementation of that method.

jodosha:
Fixed by #130

Issue #127, opened 2016-03-21
URI.extract is obsolete


jodosha:
`Hanami::Utils::Escape.url` is using `URI.extract`, which is deprecated. We need to update the implementation of that method.

jodosha:
Fixed by #130

PR #130, opened 2016-03-21
Removing warnings for URI.extract and URI.unescape because these methods are now obsolete

rogeriozambon:
Here are the warnings:

```
../lib/hanami/utils/escape.rb:514:in `url': warning: URI.extract is obsolete
../lib/hanami/utils/escape.rb:515:in `url': warning: URI.unescape is obsolete
```

This solves the problems reported in #127 and #128.


joneslee85:
:+1:

jodosha:
@rogeriozambon Thank you!

PR #131, opened 2016-03-23
Added tests to BasicObject#respond_to_missing? and String#singularize

rogeriozambon:
Hey guys, can you check if it's relevant? Just adding tests to:

```ruby
# Hanami::Utils::BasicObject

def respond_to_missing?(method_name, include_all)
  ::Kernel.raise ::NotImplementedError
end
```

and

```ruby
# Hanami::Utils::String

def singularize
  self.class.new Inflector.singularize(self)
end
```

### DIFF
--- a/test/basic_object_test.rb
+++ b/test/basic_object_test.rb
@@ -1,0 +1,25 @@
+require 'test_helper'
+require 'hanami/utils/basic_object'
+
+class TestClass < Hanami::Utils::BasicObject
+end
+
+describe Hanami::Utils::BasicObject do
+  describe '#respond_to_missing?' do
+    it 'raises an exception if respond_to? method is not implemented' do
+      -> {
+        TestClass.new.respond_to?(:no_existing_method)
+      }.must_raise(NotImplementedError)
+    end
+
+    it 'must return true if respond_to? method was implemented' do
+      TestCase = Class.new(TestClass) do
+        def respond_to?(method_name, include_all = false)
+          true
+        end
+      end
+
+      assert TestCase.new.respond_to?(:no_existing_method)
+    end
+  end
+end
--- a/test/basic_object_test.rb
+++ b/test/basic_object_test.rb
@@ -12,7 +12,7 @@ describe Hanami::Utils::BasicObject do
       }.must_raise(NotImplementedError)
     end
 
-    it 'must return true if respond_to? method was implemented' do
+    it 'returns true given respond_to? method was implemented' do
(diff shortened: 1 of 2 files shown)


rogeriozambon:
Maybe that build should be restarted. Anyone here have the power (:laughing:) to do this?

test/basic_object_test.rb
```diff
@@ -0,0 +1,25 @@
...
+
+describe Hanami::Utils::BasicObject do
+  describe '#respond_to_missing?' do
+    it 'raises an exception if respond_to? method is not implemented' do
+      -> {
+        TestClass.new.respond_to?(:no_existing_method)
+      }.must_raise(NotImplementedError)
+    end
+
+    it 'must return true if respond_to? method was implemented' do
```
joneslee85: s/must return/returns

I still prefer another `it` for stating context `given respond_to? method was implemented`

test/string_test.rb
```diff
@@ -273,6 +273,29 @@
     end
   end
 
+  describe '#singularize' do
+    before do
+      @singular, @plural = *TEST_SINGULARS.to_a.sample
```
joneslee85: not a big fan of testing against a moving random value, can we do it on fixed value instead?

test/basic_object_test.rb
rogeriozambon: @joneslee85, so your suggestion is: `returns true given respond_to? method was implemented`?

test/string_test.rb
rogeriozambon: @joneslee85 I agree with you, it is strange. But, considering that we've a bunch of fixtures, it doesn't makes more sense to use them?

jodosha: @rogeriozambon @joneslee85 Better random values for testing. We want to avoid to design tests around well known values.

test/basic_object_test.rb
jodosha: @rogeriozambon Yes, @joneslee85 makes sense :+1:

jodosha:
@rogeriozambon Thanks for this PR. Apart for [this minor comment](https://github.com/hanami/utils/pull/131/files#r57101474), it's :+1: for me :+1:

jodosha:
(I restarted the failing build)

rogeriozambon:
I've changed the test description as requested.

jodosha:
@rogeriozambon Cool, thank you! :+1:

PR #134, opened 2016-03-26
Improve single quote consistency for string_test

duykhoa:
Improve the present of `string_test`, use single quote (') whenever possible


jodosha:
@duykhoa Thank you for this PR, merging! :+1:

PR #135, opened 2016-04-07
Add missing tests for Kernel#numeric? method

davydovanton:
Reference https://github.com/hanami/utils/commit/57f6be767b917ff494476c25332e7cae501d904d

/cc @hanami/contributors @hanami/core-team

### DIFF
--- a/lib/hanami/utils/kernel.rb
+++ b/lib/hanami/utils/kernel.rb
@@ -1010,7 +1010,7 @@ module Hanami
       # @since x.x.x
       # @api private
       def self.numeric?(arg)
-        arg.to_s.match(NUMERIC_MATCHER)
+        !!arg.to_s.match(NUMERIC_MATCHER)
       end
 
       # Returns the most useful type error possible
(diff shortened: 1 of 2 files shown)


jodosha:
@davydovanton Hey, thanks for these additional tests. This method is indirectly tested via `.Integer` and `.Float`, which use it.

I don't mind to add extra tests for it, since now it's used outside of this library. This PR helped me to spot one behavior to fix: the method doesn't return a boolean result. It returns `nil`, instead of `false` and `MatchData` instance instead of `true`.

Can you please make sure that it will work as expected? Thank you very much.

davydovanton:
@jodosha YES! When I wrote tests I asked myself why this method return numeric value. I can fix this, but I think that this changes make method more slow.

jodosha:
@davydovanton You're right. The difference isn't huge for MRI 2.2, 2.3, but it is for JRuby 9k. See this benchmark: https://gist.github.com/jodosha/db1cf2696e6a4b84a95a7ab4853c5a92

davydovanton:
hm, interesting results. I was hoping that the result would be worse 😕
I think that this difference is very small that's why I prefer select more idiomatically alternative.

davydovanton:
@jodosha done

jodosha:
@davydovanton Thank you, merged!

davydovanton:
Thanks for merge :+1:

PR #136, opened 2016-04-15
Fix singularized for 'price' word

davydovanton:
Reference to https://github.com/hanami/router/issues/103

/cc @hanami/core-team and @hanami/contributors

### DIFF
--- a/lib/hanami/utils/inflector.rb
+++ b/lib/hanami/utils/inflector.rb
@@ -228,6 +228,7 @@ module Hanami
         # regressions
         # https://github.com/hanami/utils/issues/106
         'album'        => 'albums',
+        'price'        => 'prices',
       })
 
       # Irregular rules for singulars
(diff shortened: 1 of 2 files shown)


habutre:
@davydovanton I the prices isn't an irregular plural, it just gain a "s" being a simple plural.
The bug here, its on highlighted lines below on lib/hanami/utils/inflector.rb file:

Pluralize should treat only "mouse|louse|#other irregular plurals here#", including ICE will cause simple plural be incorrectly pluralized.
```
#pluralize
365 when /\A(.*)(ouse|#{ ICE })\z/
366          $1 + ICE
```

Here the lines 406,407 seems to be OK, but 418,419.
```
#singularize
406 when /\A(.*)#{ICE}\z/
407           $1 + OUSE
...
418 when /\A(.*)#{OUSE}\z/
419           $1 + ICE
```

My fix would be remove ICE constant from pluralize method and remove lines 418, 419 from singularize method. Make sense?

PR #137, opened 2016-04-28
Allow deep indifferent access on Hanami::Utils::Attributes

Erol:
Address https://github.com/hanami/controller/issues/70 by making `Hanami::Utils::Attribute#get` return `Hanami::Utils::Attributes` for hash-like attribute values.

### DIFF
--- a/lib/hanami/utils/attributes.rb
+++ b/lib/hanami/utils/attributes.rb
@@ -66,7 +66,16 @@ module Hanami
       #   attributes[:unknown]      # => nil
       #   attributes['unknown']     # => nil
       def get(attribute)
-        @attributes[attribute.to_s]
+        value = @attributes
+
+        keys = attribute.to_s.split('.')
+        keys.each do |key|
+          break unless value
+
+          value = value[key]
+        end
+
+        value.kind_of?(Hash) ? self.class.new(value) : value
       end
 
       # @since 0.3.4
--- a/test/attributes_test.rb
+++ b/test/attributes_test.rb
@@ -51,6 +51,19 @@ describe Hanami::Utils::Attributes do
       attributes.get('23').must_equal 'foo'
     end
 
+    it 'allows deep indifferent access' do
+      attributes = Hanami::Utils::Attributes.new(foo: {baz: 'bar'})
+      attributes.get(:foo).get(:baz).must_equal 'bar'
+      attributes.get(:foo).get('baz').must_equal 'bar'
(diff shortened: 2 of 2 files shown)


test/attributes_test.rb
```diff
@@ -51,6 +51,14 @@ def to_h
       attributes.get('23').must_equal 'foo'
     end
 
+    it 'allows deep indifferent access' do
+      attributes = Hanami::Utils::Attributes.new(foo: {baz: 'bar'})
+      attributes.get(:foo).get(:baz).must_equal 'bar'
```
joneslee85: good one, can we test against the nested string syntax instead? eg: `params.get('address.street')`

Erol: Oh yeah, I forgot to test that one. New commit coming up.

Erol:
How about this?

joneslee85:
looks good, the test fails for other unrelated reason. Would you be able to have a look at that too?

lib/hanami/utils/attributes.rb
```diff
@@ -66,7 +66,16 @@ def initialize(hash = {})
...
       def get(attribute)
-        @attributes[attribute.to_s]
+        value = @attributes
+
+        keys = attribute.to_s.split('.')
+        keys.each do |key|
+          break unless value
+
+          value = value[key]
+        end
```
davydovanton: What do you think, maybe is better to get only last key from `split` instead unnecessary enumeration? I. e.:
```ruby
value = @attributes

key = attribute.to_s.split(?.).last
value = value[key] if value
```

Erol:
@joneslee85: Can't reproduce the error on my local, but sure I'll also take a look at it.

lib/hanami/utils/attributes.rb
Erol: We wouldn't be able to cover deep access with that, like for example:

```
attributes = Hanami::Utils::Attributes.new({address: {street: 'Street', city: 'City'}})
attributes.get("address.street")
```

davydovanton: yep, you're right!
Thanks for explanation! 👍

Erol: 👍

Erol:
I squashed my commits and Travis already passed. Maybe a brittle test for file permissions?

Erol:
Thanks for the merge!

PR #139, opened 2016-05-16
Fixing the plural of words ending by *ice

habutre:
Including mouse and louse on irregular plurals
Including policie as uncountable plural
Words like price, advice was not being pluralized as expected

I take the freedom to open this PR aim to fix this issues, please take a look on that [PR](https://github.com/hanami/utils/pull/136) where we start such discussion.


joneslee85:
The PR is good but I am wondering should we take in more changes to the inflection mapping as there are many exceptions. Should we leave it to the user instead? Thought?

rogerleite:
Maybe [ActiveSupport inflections](https://github.com/rails/rails/blob/master/activesupport/lib/active_support/inflections.rb#L10) can help to review it. 

Great job @habutre 👍

jodosha:
@habutre Thank you!

PR #142, opened 2016-05-30
Create wrapper for json module

davydovanton:
/cc @hanami/contributors, @joneslee85 and @jodosha

### DIFF
--- a/lib/hanami/logger.rb
+++ b/lib/hanami/logger.rb
@@ -1,6 +1,7 @@
 require 'json'
 require 'logger'
 require 'hanami/utils/string'
+require 'hanami/utils/json'
 
 module Hanami
   # Hanami logger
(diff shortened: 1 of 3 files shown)


lib/hanami/logger.rb
```diff
@@ -156,7 +157,7 @@ class JSONFormatter < Formatter
       # @since x.x.x
       # @api private
       def _format(hash)
-        JSON.generate(hash)
+        Hanami::Utils::Json.dump(hash)
```
joneslee85: should upcase the Json keyword?

joneslee85:
LGTM

lib/hanami/logger.rb
davydovanton: I think no. I want to use hanami json wrapper here, that's why I use downcased `json` keyword

jodosha:
@davydovanton Thank you. I just merged this!

davydovanton:
Thanks for merge! :tada:

PR #143, opened 2016-05-30
Add end of line for default logger format

davydovanton:
1. Don't display empty values
2. Display new line

/cc @jodosha and @joneslee85

### DIFF
--- a/lib/hanami/logger.rb
+++ b/lib/hanami/logger.rb
@@ -103,6 +103,10 @@ module Hanami
       # @api private
       SEPARATOR = ' '.freeze
 
+      # @since x.x.x
+      # @api private
+      NEW_LINE = "\n".freeze
+
       # @since 0.5.0
       # @api private
       attr_writer :application_name
@@ -142,7 +146,7 @@ module Hanami
       # @since x.x.x
       # @api private
       def _format(hash)
-        hash.map { |k, v| "#{k}=#{v}" }.join(SEPARATOR)
+        hash.map { |k, v| "#{k}=#{v}" unless v.nil? }.join(SEPARATOR) + NEW_LINE
       end
     end
 
--- a/lib/hanami/logger.rb
+++ b/lib/hanami/logger.rb
@@ -146,7 +146,7 @@ module Hanami
       # @since x.x.x
       # @api private
       def _format(hash)
-        hash.map { |k, v| "#{k}=#{v}" unless v.nil? }.join(SEPARATOR) + NEW_LINE
+        hash.map { |k, v| "#{k}=#{v}" }.join(SEPARATOR) + NEW_LINE
       end
     end
 
--- a/lib/hanami/logger.rb
+++ b/lib/hanami/logger.rb
@@ -105,7 +105,7 @@ module Hanami
 
       # @since x.x.x
       # @api private
-      NEW_LINE = "\n".freeze
+      NEW_LINE = $/
 
       # @since 0.5.0
       # @api private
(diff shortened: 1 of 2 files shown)


lib/hanami/logger.rb
```diff
@@ -142,7 +146,7 @@ def _message_hash(message)
       # @since x.x.x
       # @api private
       def _format(hash)
-        hash.map { |k, v| "#{k}=#{v}" }.join(SEPARATOR)
+        hash.map { |k, v| "#{k}=#{v}" unless v.nil? }.join(SEPARATOR) + NEW_LINE
```
joneslee85: can we have a test to assert the case when value is nil? - I don't think it is a bad idea to show `nil` as string. What's the reason behind adding this guard?

lib/hanami/logger.rb
```diff
@@ -103,6 +103,10 @@ class Formatter < ::Logger::Formatter
       # @api private
       SEPARATOR = ' '.freeze
 
+      # @since x.x.x
+      # @api private
+      NEW_LINE = "\n".freeze
```
joneslee85: FYI, this won't work on Windows. You can use the special EOL constant `$/` which works everywhere

lib/hanami/logger.rb
jodosha: Yes, why `nil` values would be passed here? And in that case, it would be probably good to keep the key.

```ruby
{a: nil, b: 1}.map { |k, v| "#{k}=#{v}" }.join(" ")
# => "a= b=1"
```

@joneslee85 @davydovanton wdyt?

davydovanton: I think that nil value takes a lot of space in `STDOUT`, that's why I want to show only actual values.

joneslee85: I am 👍 for keeping empty value. It does take a bit more spaces but in return the format stays consistent for log consumer to play with.

jodosha: @davydovanton Do you have a practical example?

davydovanton: @jodosha  yes, check this:
```
Puma starting in single mode...
* Version 3.4.0 (ruby 2.3.0-p0), codename: Owl Bowl Brawl
* Min threads: 0, max threads: 16
* Environment: development
* Listening on tcp://localhost:2300
Use Ctrl-C to stop
app=Web severity=DEBUG time=0.00774 addr=::1  now_time=01/Jun/2016:10:26:47 +0300 request_method=GET path=  http_version=HTTP/1.1 status=200
```

and this
```
Puma starting in single mode...
* Version 3.4.0 (ruby 2.3.0-p0), codename: Owl Bowl Brawl
* Min threads: 0, max threads: 16
* Environment: development
* Listening on tcp://localhost:2300
Use Ctrl-C to stop
app=Web severity=DEBUG time=0.005107 addr=::1 user= now_time=01/Jun/2016:10:27:13 +0300 request_method=GET path= query_string= http_version=HTTP/1.1 status=200 length=
```

Do you really what to see empty `length=`, `user=` and etc?

davydovanton: But any way I can keep empty values. If it be bad we will change this 👍

davydovanton: @jodosha nope, I mean that we can change this line in any time if we need

jodosha:
@davydovanton Merging this with only the new line support. If we notice we need to remove blank values, we can do it later. Thanks! 👍

davydovanton:
@jodosha absolutely agree, thanks! 👍

PR #144, opened 2016-06-02
Extract duplicate method logic

gogainda:
Hello!

This is my attempt to increase codeclimate score for hash.rb
This code is not perfect, but befor I start improving it, I want to know your opinion @jodosha worth this PR to be working on or we can close it right away?


lib/hanami/utils/hash.rb
```diff
@@ -63,12 +63,8 @@ def initialize(hash = {}, &blk)
...
       #   hash.inspect # => {:a=>23, :b=>{:c=>["x", "y", "z"]}}
       def symbolize!
-        keys.each do |k|
-          v = delete(k)
-          v = self.class.new(v).symbolize! if v.is_a?(::Hash) || v.is_a?(self.class)
-
-          self[k.to_sym] = v
-        end
+        @hash = HashKeysConverter.new(@hash).to_symbols
+        puts "flksdjfls"
```
davydovanton: missed `puts`, yes? :)

gogainda: yep)

gogainda: btw, where can I get rubocop config which is used in codeclimate for checking?

gogainda:
```Coverage increased (+0.0004%) ``` true win-win)

**Comment by gogainda** [comment]
@jodosha np
In addition to this PR I wanted to introduce non-destructive version of stingify! and symbolize!
```
def symbolize
        duplicate = self.deep_dup

        self.class.new(convert_hash_keys_using(duplicate.to_h, :to_sym))
      end
```
Will create it in the next PR

PR #145, opened 2016-06-23
Fix typo

jooohn:
fix typo :)


jodosha:
@jooohn Thanks for this fix. 😄

PR #148, opened 2016-07-16
Add methods #successful? and #failing? to Interactor

mgrachev:
Example:
```
module Web::Controllers::Projects
  class Create
    include Web::Action

    params do
      param :project do
        param :name, presence: true
      end
    end

    def call(params)
      return unless params.valid?

      # Interactor
      create_project = CreateProject.new(params[:project]).call
      return if create_project.failure?

      redirect_to routes.projects_path
    end
  end
end
```

### DIFF
--- a/lib/hanami/interactor.rb
+++ b/lib/hanami/interactor.rb
@@ -17,7 +17,14 @@ module Hanami
       # @api private
       #
       # @see Hanami::Interactor::Result#respond_to_missing?
-      METHODS = ::Hash[initialize: true, success?: true, fail!: true, prepare!: true, errors: true, error: true].freeze
+      METHODS = ::Hash[initialize:  true,
+                       success?:    true,
+                       successful?: true,
+                       failing?:    true,
+                       fail!:       true,
+                       prepare!:    true,
+                       errors:      true,
+                       error:       true].freeze
 
       # Initialize a new result
       #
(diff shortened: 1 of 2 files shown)


joneslee85:
cc @hanami/core-team 

I am 👍 with this convenient method

jodosha:
@mgrachev What about:

```ruby
return unless result.valid?
```

Which is similar to the line at the begin of the method:

```ruby
return unless params.valid?
```

cllns:
I like this!

@jodosha Do you mean `return unless result.success?`

I prefer the implementation @mgrachev wrote. It will return `true/false` rather than `true/nil`. Any reason `true/nil` would be better?

jodosha:
@cllns Yup, sorry there was a typo.

My suggestion is to make consistent how the guards are read in that method.

```ruby
def call(params)
    return unless params.valid?

    # Interactor
    create_project = CreateProject.new(params[:project]).call
    return unless create_project.success? # This is my suggestion: it's consistent with the first line of this method.

    redirect_to routes.projects_path
end
```

What do you think?

cllns:
ah, I see. I thought you were saying `return unless create_project.success?` should be the body of the `failure?` mehod. 

Yea, that looks good in this case but I don't really have a preference either way. I think `.failure?` is a useful convenience method though, for people who use Interactors in a different way (like outside of an Action, for a non-web usage).

Honestly, I like to avoid `unless` whenever possible, so I'd prefer to write the first line as `return if params.invalid?` (if `params.invalid?` were implemented) :D

jodosha:
@mgrachev @joneslee85 @cllns I have to confess that I don't like the original choice of `#success?` because it reads bad on tests.

Minitest:

```ruby
result.must_be :success?
```

RSpec:

```ruby
expect(result).to be_success
```

I think we should improve that. My proposal is:

  1. Define `#successful?` method
  2. Make `#success?` an alias of `#successful?` for backward compatibility
  3. Introduce `#failing?` (instead of `#failure?`) as opposite of `#successful?`

What do you think?

joneslee85:
@jodosha I like that proposal, to me adjective `successful` is much better in grammar.

jodosha:
Let's go for it! 😉

mgrachev:
@jodosha Done.

jodosha:
@mgrachev Thank you very much. Merging it!

Issue #151, opened 2016-08-09
namespace issue with load_from_pattern!


pascalbetz:
For 

```ruby
class Bar
  def level
    "top"
  end
end

class Foo
  class Bar
    def level
       "nested"
    end
  end
end
```

I'd expect

```ruby
klass = Hanami::Utils::Class.load_from_pattern!('(Hanami|Foo)::Bar')
assert_equal 'nested', klass.new.level
```
to pass. But it won't as ```puts klass.new.level #=> 'top'```

jodosha:
@pascalbetz Do you think that `Utils::Class.load_from_pattern!` [should use](https://github.com/hanami/utils/blob/a958dd528cc4b8e7ccb7cdd4b96f3aac7b406526/lib/hanami/utils/class.rb#L115) `const_get(name, false)`?

pascalbetz:
Yes. Perhaps you can check out the PR. I've added 3 test cases. Two of which work plus one that fails with current implementation. All of them pass with changed implementation.

jodosha:
@pascalbetz I noticed the PR after my question. I'm looking at it and checking if all the other frameworks are affected from that proposed change. Thanks!

PR #152, opened 2016-08-09
Making sure that only current namespace is checked when loading classes.

pascalbetz:
Fixes https://github.com/hanami/utils/issues/151

Honestly i'm not sure about the side-effects of this to current hanami. But I think the behavior described in issue 151 is a bug.


jodosha:
@pascalbetz Everything's fine 👍

PR #153, opened 2016-08-10
Make Utils::BasicObject pretty printable

jodosha:
Ref https://github.com/hanami/hanami/issues/629


joneslee85:
:+1:

cllns:
Nice! I never knew the interface was a `pretty_print` method. Makes sense :)

PR #160, opened 2016-10-14
Introducing Hanami::Events

jodosha:
#### Subscribe With A Proc

``` ruby
require 'concurrent'
require 'hanami/events'
require 'ostruct'

Hanami::Events.subscribe('signup') do |event|
  puts "New signup: #{event[:user].email}"
end

user = OpenStruct.new(id: 1, email: "user@hanamirb.org")
Hanami::Events.broadcast('signup', user: user)

  # => "New signup: user@hanamirb.org
```
#### Subscribe With An object

``` ruby
require 'concurrent'
require 'hanami/events'
require 'ostruct'

class WelcomeMailer
  def call(event)
    puts "New welcome message for: #{event[:user].email}"
  end
end

Hanami::Events.subscribe('signup', WelcomeMailer.new)

user = OpenStruct.new(id: 1, email: "user@hanamirb.org")
Hanami::Events.broadcast('signup', user: user)

  # => "New welcome message for: user@hanamirb.org
```
#### Include Mixin

``` ruby
require 'concurrent'
require 'hanami/events'
require 'ostruct'

class Signup
  include Hanami::Events

  def call(params)
    user = OpenStruct.new(params)
    broadcast('signup', user: user)
  end
end

Hanami::Events.subscribe('signup') do |event|
  puts "New signup: #{event[:user].email}"
end

Signup.new.call(email: "user@hanamirb.org")
  # => "New signup: user@hanamirb.org
```

---
### Useful for

This can be useful for:
- Automatic logging https://github.com/hanami/hanami/issues/484
- Move mailers to apps https://github.com/hanami/hanami/issues/648
- Async/distributed jobs
#### Automatic Logging

All the Hanami frameworks can broadcast events for operations that we want to automatically log. Things like action response, rendered templates/layouts, database queries/commands. The logger can subscribe to these events and emit a new entry.
#### Move mailers to apps

See https://github.com/hanami/hanami/issues/648#issuecomment-253855250
#### Async/Distributed Jobs

The proposed implementation ships with a memory backend as dispatcher for the events. Which is fine for basic setups, but it has the downside of being volatile.

In the future, third-party devs can implement a backend for async/distributed mechanisms like Sidekiq or Apache Kafka.

---

@hanami/core-team @hanami/contributors Please review. 💚

### DIFF
--- a/lib/hanami/events.rb
+++ b/lib/hanami/events.rb
@@ -1,0 +1,180 @@
+require 'hanami/utils/class_attribute'
+begin
+  require 'concurrent'
+rescue LoadError
+  raise "Cannot find `concurrent-ruby' gem, please install it and retry."
+end
+
+module Hanami
+  # Event framework for Hanami
+  #
+  # @since x.x.x
+  module Events
+    # Backends for Hanami::Events
+    #
+    # @since x.x.x
+    module Backends
+      # Memory backend for Hanami::Events
+      #
+      # @since x.x.x
+      # @api private
+      class Memory
+        # Instantiate a new memory backend
+        #
+        # @return [Hanami::Events::Backends::Memory] a new instance
+        #
+        # @since x.x.x
+        # @api private
+        def initialize
+          @topics = Concurrent::Map.new { |h, k| h[k] = Concurrent::Array.new }
+        end
+
+        # Subscribe to a topic
+        #
+        # @param topic [String,Symbol] the topic
+        # @param subscriber [#call] the subscriber
+        #
+        # @since x.x.x
+        # @api private
+        def subscribe(topic, subscriber)
+          topics[topic.to_sym].push(subscriber)
+        end
+
+        # Broadcast an event for a given topic
+        #
+        # @param topic [String,Symbol] the topic
+        # @param event [Hanami::Events::Event] the event
+        #
+        # @since x.x.x
+        # @api private
+        def broadcast(topic, event)
+          topics[topic.to_sym].each do |subscriber|
+            subscriber.call(event)
+          end
+        end
+
+        private
+
+        # @since x.x.x
+        # @api private
+        attr_reader :topics
+      end
+    end
+
+    # Broadcasted Event
+    #
+    # @since x.x.x
+    class Event
+      # Instantiate a new event
+      #
+      # @param [#to_hash,#to_h] the payload
+      #
+      # @return [Hanami::Events::Event] a new instance
+      #
+      # @since x.x.x
+      # @api private
+      def initialize(payload)
+        @payload = Utils::Hash.new(payload).symbolize!.freeze
+      end
+
+      # Return the value associated to the given key
+      #
+      # @param [Symbol] the key
+      #
+      # @return [Object,NilClass] the value, if present
+      #
+      # @since x.x.x
+      def [](key)
+        @payload[key]
+      end
+    end
+
+    include Utils::ClassAttribute
+
+    # Support multiple backend for the future.
+    #
+    # For now the only backend is Hanami::Events::Backends::Memory
+    #
+    # @since x.x.x
+    # @api private
+    class_attribute :backend
+    self.backend = Backends::Memory.new
+
+    # Subscribe to a topic
+    #
+    # @param topic [String,Symbol] the topic
+    # @param subscriber [#call] the subscriber (object)
+    # @param blk [Proc] the subscriber (proc)
+    #
+    # @since x.x.x
+    #
+    # @example Subscribe With An Object
+    #   require 'hanami/events'
+    #
+    #   class WelcomeMailer
+    #     def call(event)
+    #       # do something with event
+    #     end
+    #   end
+    #
+    #   Hanami::Events.subscribe('welcome', WelcomeMailer.new)
+    #
+    # @example Subscribe With A Proc
+    #   require 'hanami/events'
+    #
+    #   Hanami::Events.subscribe('welcome') do |event|
+    #     # do something with event
+    #   end
+    def self.subscribe(topic, subscriber = nil, &blk)
+      backend.subscribe(topic, subscriber || blk)
+    end
+
+    # Broadcast an event for a given topic
+    #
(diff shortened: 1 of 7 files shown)


mereghost:
Heya @jodosha, hope to find you well.

Is this mainly targeted at Hanami internals? If not why are we offering it when there are awesome libs already around for the developer to choose?

I mean event driven approaches are full of little gotchas that maybe the team will get overwhelmed by feature requests like retries, delayed deliveries, dead lettering etc.

lib/hanami/events.rb
```diff
@@ -0,0 +1,180 @@
...
+      # @api private
+      class Memory
+        # Instantiate a new memory backend
+        #
+        # @return [Hanami::Events::Backends::Memory] a new instance
+        #
+        # @since x.x.x
+        # @api private
+        def initialize
+          @topics = Concurrent::Map.new { |h, k| h[k] = Concurrent::Array.new }
```
AlfonsoUceda: @jodosha isn't it enough to use a Mustex here?

lib/hanami/events.rb
```diff
@@ -0,0 +1,180 @@
...
+      #
+      # @since x.x.x
+      def [](key)
+        @payload[key]
+      end
+    end
+
+    include Utils::ClassAttribute
+
+    # Support multiple backend for the future.
```
AlfonsoUceda: Backend => backends?

lib/hanami/events.rb
```diff
@@ -0,0 +1,180 @@
...
+        # @since x.x.x
+        # @api private
+        attr_reader :topics
+      end
+    end
+
+    # Broadcasted Event
+    #
+    # @since x.x.x
+    class Event
```
AlfonsoUceda: Why not move to its own class?

lib/hanami/events.rb
```diff
@@ -0,0 +1,180 @@
...
+      # Instantiate a new event
+      #
+      # @param [#to_hash,#to_h] the payload
+      #
+      # @return [Hanami::Events::Event] a new instance
+      #
+      # @since x.x.x
+      # @api private
+      def initialize(payload)
+        @payload = Utils::Hash.new(payload).symbolize!.freeze
```
AlfonsoUceda: @pascalbetz do you have in mind other thing? I think a hash is more flexible.

pascalbetz: Just whatever I want to broadcast. So instead of 

```ruby
event = Event.new(email: 'pascal@something.com')
email = event.payload[:email]
```

I could do

```ruby
event = Event.new('pascal@something.com')
email = event.payload
```

AlfonsoUceda:
@jodosha good PR I was wondering if we could add default data to the event like a timestamp for example. What do you think?

Another point, what do you think to add support to regex when subscribe? I mean sometimes we want catch all events related to one topic, for exmaple we could have several events in hanami model, maybe we want to catch all of them.

pascalbetz:
@AlfonsoUceda, @jodosha are we reinventing the wheel? Isn't there an existing solution that would fit? Wisper comes to mind. But surely there are others.

TiteiKo:
Just a thought here... I don't think I'm mistaken in saying that the memory backend will be process (not thread, but process) specific, and so server specific. It should be mentioned somewhere in the docs/comments, as to not take by surprise someone who would not have dig in on how it works.

jodosha:
Folks, thanks for all your comments. The intent here is not to reinvent the wheel, but to offer a **basic**, **internal** and **quick** way to broadcast events. Let me say it's a MVP to _"unlock"_ the features described above.

---

I'm sure that `wisper` is better, but I wanted to avoid to expose extra APIs that we're not sure we want to maintain for long time. Think of `Action#flash`, it's marked as private/experimental, but people are already using it.

This could happen with new libraries that we may want to introduce. This PR adds a bare minimum public API (`#broadcast`) that we'll probably be there, regardless the internal implementation.

---

At this point we're not interested into retry, async, timestamps, unsubscribe, persistent queues, etc.. but to ship an event broadcaster ASAP. The only modification is the @TiteiKo suggestion to reduce the scope of the broadcast to the current thread.

If you think this is possible with the proposed API/implementation, please vote here thumbs up/down. I'm fine to close this as long we have another PR ready soon.

---

Don't get me wrong, this proposal is demanded bye the urgency, not by NIH. If you will to help with better solutions (including `wisper`), please do. 😄

mereghost:
@jodosha that being the case you have my 👍 . I'm a fan of event-driven approaches and having something like that builtin for internals can help a lot with decoupling while keeping the dependencies at the bare minimum.

lib/hanami/events.rb
mereghost: Not sure if advantageous in any given situation tho, @pascalbetz. Can you show me an use case that makes passing anything-but-a-hash relevant?

pascalbetz: @mereghost there is no need or advantage to force payload to be hash. Just let it be whatever the user wants it to be. 
Besides that I don't see a reason why I should wrap some of the following into a hash, just to unwrap it afterwards:

* List of users
* some id
* username
* email address
* ...

AlfonsoUceda:
@jodosha 👍 if we need more in the future we can iterate in future steps

TiteiKo:
@jodosha 👍

theCrab:
👍

AlfonsoUceda:
@jodosha could yo rebase master in this PR? thanks man ;)

jodosha:
I'm reconsidering my position. This proposal wasn't included in Hanami v0.9.0 because it was controversial.

---

My proposal is to:

  * Use the same proposed interface for `subscribe`/`broadcast`
  * Change the engine in favor of `wisper` (as a soft dependency for `hanami-utils`)

Latest `wisper` version is `2.0.0.rc1`, released almost two years ago.
@pascalbetz because you have contributed to that project, can you please check with @krisleech if we can help him to finalize a `2.0.0`?

---

What do you folks think?

jodosha:
I'm closing this. We can discuss about this as a proposal after 1.0

PR #161, opened 2016-10-14
Hanami::Utils::Hash#symbolize! now handles array

lucasallan:
Fixes https://github.com/hanami/hanami/issues/653

### DIFF
--- a/lib/hanami/utils/hash.rb
+++ b/lib/hanami/utils/hash.rb
@@ -67,6 +67,17 @@ module Hanami
           v = delete(k)
           v = self.class.new(v).symbolize! if v.respond_to?(:to_hash)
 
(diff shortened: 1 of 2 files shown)


cllns:
Should we see if it's Enumerable rather than seeing if it's an Array?

jodosha:
@lucasallan Thanks for this PR. I cleaned up a bit, according to the changes introduced by #162

jodosha:
@cllns The implementation now uses `Enumerable` instead of `Array`.

jodosha:
I had to revert this and remove from the release because it breaks `Hanami::Entity`. https://travis-ci.org/hanami/model/builds/185408316

Ref 3f8ebf8

Issue #159, opened 2016-10-13
Interactor does not allow to expose hashes with non-symbolizable keys


alexd16:
Hi,

I was trying to use the Hanami::Interactor in a rails project and I came across an issue where I cannot expose a hash that does not have symbolizable keys (to_sym not defined).

Example: 

``` ruby
class Mapping
  include Hanami::Interactor

  expose :mapping

  def call
    @mapping = { 1 => 'Hello' }
  end
end

Mapping.new.call
# NoMethodError: undefined method `to_sym' for 1:Fixnum
```

From what I see the issue is that the exposures in the Interactor is stored in a Utils::Hash and it calls simbolize! here: https://github.com/hanami/utils/blob/master/lib/hanami/interactor.rb#L136.

This will try to convert every keys in the exposures hash and any nested hash to a symbol which will blow up in this case because Fixnum does not have the to_sym method.

I was wondering if this is the intended behaviour and If so if there is work around to implement something like this?

Thank you

jodosha:
@alexd16 Hi! Yes, that isn't the expected way to use it. The returning value of `#call` is meant to be a symbolized hash.

May I ask you what's your real world use case? Thanks

alexd16:
Hi @jodosha! Thanks for the response.

About my use case, I need to render an excel type grid in the frontend. 
This grid is used to join two models and assign a certain float value.

So the easiest way I found to fill the initial values of the grid was to build an hash with ids of one of the models mapping to the ids of the other model with the value that should appear in the respective cell. For example:
```ruby
mapping: {
  1: {
    5: 2.0,
    7: 1.0
  },
  10: {
    1: 0.5,
    12: 1.4
  }
}
```

So I was trying to use an interactor to build the hash and expose it to the controller.

Not sure if this is a scenario where I should use an Interactor or not.
What do you think?

If something is not clear let me know. Thks

PS: Sorry to bother you with this.

jodosha:
@alexd16 

> PS: Sorry to bother you with this.

No worries at all.

---

Now I see your problem. It's because Interactor uses `Utils::Hash#symbolize!` for result. Unfortunately that `Utils::Hash` method deep symbolizes all the keys, which is unexpected.

The intent here is to just symbolize the keys of result:

```ruby
Hanami::Interactor::Result.new("foo" => { "bar" => "baz" })
```

In the example above only `"foo"` should be symbolized, not the nested hash.

alexd16:
[...]

PR #162, opened 2016-10-19
Hanami::Interactor Only symbolize keys of result payload

alexd16:
Fixes: https://github.com/hanami/utils/issues/159

/cc @jodosha

### DIFF
--- a/lib/hanami/utils/hash.rb
+++ b/lib/hanami/utils/hash.rb
@@ -48,7 +48,7 @@ module Hanami
         @hash.default_proc = blk
       end
 
-      # Convert in-place all the keys to Symbol instances, nested hashes are converted too.
+      # Convert in-place all the keys to Symbol instances. Nested hashes are converted too if deep argument is truthy.
       #
       # @return [Hash] self
       #
@@ -62,11 +62,10 @@ module Hanami
       #
       #   hash.keys    # => [:a, :b]
       #   hash.inspect # => {:a=>23, :b=>{:c=>["x", "y", "z"]}}
-      def symbolize!
+      def symbolize!(deep = true)
         keys.each do |k|
           v = delete(k)
-          v = self.class.new(v).symbolize! if v.respond_to?(:to_hash)
-
+          v = self.class.new(v).symbolize! if v.respond_to?(:to_hash) && deep
           self[k.to_sym] = v
         end
 
--- a/lib/hanami/utils/hash.rb
+++ b/lib/hanami/utils/hash.rb
@@ -48,7 +48,7 @@ module Hanami
         @hash.default_proc = blk
       end
 
-      # Convert in-place all the keys to Symbol instances. Nested hashes are converted too if deep argument is truthy.
+      # Convert in-place all the keys to Symbol instances.
       #
       # @return [Hash] self
       #
@@ -61,11 +61,34 @@ module Hanami
       #   hash.symbolize!
       #
       #   hash.keys    # => [:a, :b]
-      #   hash.inspect # => {:a=>23, :b=>{:c=>["x", "y", "z"]}}
-      def symbolize!(deep: true)
+      #   hash.inspect # => { :a => 23, :b => { 'c' => ["x", "y", "z"] } }
+      def symbolize!
         keys.each do |k|
           v = delete(k)
-          v = self.class.new(v).symbolize! if v.respond_to?(:to_hash) && deep
+          self[k.to_sym] = v
+        end
+
+        self
+      end
+
+      # Convert in-place all the keys to Symbol instances, nested hashes are converted too.
+      #
+      # @return [Hash] self
+      #
+      # @since 0.1.0
+      #
+      # @example
+      #   require 'hanami/utils/hash'
+      #
+      #   hash = Hanami::Utils::Hash.new 'a' => 23, 'b' => { 'c' => ['x','y','z'] }
+      #   hash.deep_symbolize!
+      #
+      #   hash.keys    # => [:a, :b]
+      #   hash.inspect # => {:a=>23, :b=>{:c=>["x", "y", "z"]}}
+      def deep_symbolize!
+        keys.each do |k|
+          v = delete(k)
+          v = self.class.new(v).symbolize! if v.respond_to?(:to_hash)
           self[k.to_sym] = v
         end
 
--- a/lib/hanami/utils/hash.rb
+++ b/lib/hanami/utils/hash.rb
@@ -62,7 +62,7 @@ module Hanami
       #
       #   hash.keys    # => [:a, :b]
       #   hash.inspect # => {:a=>23, :b=>{:c=>["x", "y", "z"]}}
-      def symbolize!(deep = true)
+      def symbolize!(deep: true)
         keys.each do |k|
           v = delete(k)
           v = self.class.new(v).symbolize! if v.respond_to?(:to_hash) && deep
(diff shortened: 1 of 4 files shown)


lib/hanami/utils/hash.rb
```diff
@@ -62,11 +62,10 @@ def initialize(hash = {}, &blk)
       #
       #   hash.keys    # => [:a, :b]
       #   hash.inspect # => {:a=>23, :b=>{:c=>["x", "y", "z"]}}
-      def symbolize!
+      def symbolize!(deep = true)
```
alexd16: Yeah, that makes sense. Will update the PR.

test/hash_test.rb
```diff
@@ -33,29 +33,40 @@
...
-      nested = Hanami::Utils::Hash.new('metadata' => WrappingHash.new('coverage' => 100))
-      nested.symbolize!
+    describe 'with deep false' do
+      it 'does not simbolize nested hashes' do
+        hash = Hanami::Utils::Hash.new('nested' => { 'key' => 'value' })
+        hash.symbolize!(deep: false)
 
-      nested[:metadata].must_be_kind_of Hanami::Utils::Hash
-      nested[:metadata][:coverage].must_equal(100)
+        hash[:nested]['key'].must_equal('value')
```
joneslee85: What do you think about checking the `keys` of `hash[:nested]` instead? To me it is clearer in intent as we do only assert that the key is not symoblized

test/interactor_test.rb
```diff
@@ -412,6 +412,11 @@ def initialize(_user, params)
       result.b.must_equal 2
     end
 
+    it 'returns hash values passed in the payload' do
```
joneslee85: What do you think about changing the description to `it "returns a hash value that is symbolized shallowly"`?

lib/hanami/utils/hash.rb
```diff
@@ -48,7 +48,7 @@ def initialize(hash = {}, &blk)
         @hash.default_proc = blk
       end
 
-      # Convert in-place all the keys to Symbol instances, nested hashes are converted too.
+      # Convert in-place all the keys to Symbol instances. Nested hashes are converted too if deep argument is truthy.
```
joneslee85: What do you think about changing the wording to `, optional nested hashes`

alexd16: Like this: Convert in-place all the keys to Symbol instances, optionally nested hashes are converted too if deep argument is truthy. ?

test/interactor_test.rb
alexd16: But the value is not symbolized, I am not sure if that is very clear. This test is to verify that the values of the Hash passed as an argument are not changed.

For example: Currently if we have
result = Interactor::Result.new({ m: { "key" => "value" } })
result.m => { key: "value" }
and it should be
result.m => { "key" => "value" }

To be honest I am not sure if we need this additional test. We already have a test before this one that has:

```ruby
    it 'returns all the values passed in the payload' do
      result = Hanami::Interactor::Result.new(a: 1, b: 2)
      result.a.must_equal 1
      result.b.must_equal 2
    end
``` 

Given that description the current behaviour is already failing it. Our test code is just not verifying it.
Maybe we could change this test to contemplate more cases passing a more complex hash in the constructor.

What do you think?

test/hash_test.rb
alexd16: Yeah that makes sense. Will update the PR, thank you

jodosha:
@alexd16 Hi and thanks for this PR.

Can we separate `Utils::Hash#symbolize!` from `Utils::Hash#deep_symbolize!`? Thanks.

alexd16:
@jodosha  Yeah sure, not problem. 
But that could break current usage of the method symbolize! right? Is that ok?
I could create a separate method shallow_symbolize! to avoid this. 

What do you prefer?

jodosha:
@alexd16 

> But that could break the current usage of the method symbolize! if someone is using the method expecting that it does a deep symbolize right?

Yes it's true, but we're still under 1.0, this is the right time to do this change as we'll hit that milestone soon.

lib/hanami/utils/hash.rb
joneslee85: yeah

alexd16:
@jodosha Sorry I didn't pick up the PR for some time. 

I have another question about separating symbolize! method into two. 

The current implementation searches every value that responds to the to_hash method and wraps it into an instance of the Utils::Hash class. 

I think this was done only to be able to call the symbolize! method for those hashes, and since we are no longer doing that I was wondering if we still need to do it.

What do you think?

lib/hanami/utils/hash.rb
```diff
@@ -61,11 +61,34 @@ def initialize(hash = {}, &blk)
...
+      def symbolize!
+        keys.each do |k|
+          v = delete(k)
+          self[k.to_sym] = v
+        end
+
+        self
+      end
+
+      # Convert in-place all the keys to Symbol instances, nested hashes are converted too.
```
cllns: What about nested arrays of hashes? Can you add a test for that please? 🌸

cllns: What about nested arrays of hashes? Can you add a test for that please? 🌸

alexd16: For the deep_symbolize! method I was keeping the behaviour that exists currently in master. From what I see it does not symbolize nested array of hashes, but I noticed that there is a PR to support that thought: https://github.com/hanami/utils/pull/161

Should I add that behaviour in this PR?

AlfonsoUceda:
@alexd16 thanks for the PR, could you add a test if the new method symbolize deep hashed with 3 or 4 levels? thanks ;)

AlfonsoUceda:
to me it's ok, I think it's ready to be merged cc @jodosha

jodosha:
@alexd16 Thanks for this fix. LGTM 👍 

@alexd16 @AlfonsoUceda We should verify that all the other Hanami gems work fine with this change.

alexd16:
@jodosha Happy to help. Glad I could have my first open-source contribution accepted 😄 

About verifying that this works with the other Hanami gems, how can I do that? Not sure how to proceed or If I need to do anything else.

AlfonsoUceda:
@alexd16 you should check every hanami component using hanami/utils, and in the Gemfile point to hanami/utils against your fork and branch and then run the tests. If you have problems running tests you could create a PR and delegate the work to travis.

mereghost:
And this will need to be on the upgrade notes as this breaks the behavior set in #116 since it split the methods.

AlfonsoUceda:
@alexd16 could you update your fork with the last changes in upstream? your fork is 0.8.0 version and we are no in 0.9.1. Your fork needs to be updated for testing this branch.

Thanks!

alexd16:
@AlfonsoUceda Should I update with master or 0.9.1?

AlfonsoUceda:
@alexd16 with master I prefer so you'll have your fork up to date.

alexd16:
@AlfonsoUceda Done.

Btw I ran the tests of the other Hanami gems locally and the hanami-model and hanami-controller have failing specs with this change.

I am just not sure how to fix those without this being merged.

jodosha:
@alexd16 Thank you. I finally merged this. 👍

oana-sipos:
Thanks @alexd16 for your contribution and for giving me the idea to write a small piece of documentation on how to proceed when you first contribute with code ;) very valid question! 👍

PR #166, opened 2016-10-27
New rules for words ending in Z and ZZES

### LINKED ISSUE #165: Wrong inflections on irregular words when prefixed
I have a route resource `:satisfaction_quizzes` 

Generated route for `:show` is named `satisfaction_quizze` instead of `satisfaction_quiz`

I'll just add an exception to the inflector, but is having a prefix to an irregular word a common enough occurence so that it should be handled by the inflector?

AlfonsoUceda:
This PR fix #165 

Add rule for words ending in Z they are pluralized with word + ZES.
Add rule for words ending in ZZES they are singularized with word - ZES.

cc @jodosha @TiteiKo

### DIFF
--- a/lib/hanami/utils/inflector.rb
+++ b/lib/hanami/utils/inflector.rb
@@ -191,6 +191,14 @@ module Hanami
       # @api private
       Y    = 'y'.freeze
 
+      # @since x.x.x
+      # @api private
+      Z    = 'z'.freeze
+
+      # @since x.x.x
+      # @api private
+      ZZES = 'zzes'.freeze
+
       include Utils::ClassAttribute
 
       # Irregular rules for plurals
@@ -211,7 +219,6 @@ module Hanami
         'mouse'       => 'mice',
         'ox'          => 'oxen',
         'person'      => 'people',
-        'quiz'        => 'quizzes',
         'sex'         => 'sexes',
         'testis'      => 'testes',
         'tooth'       => 'teeth',
@@ -253,7 +260,6 @@ module Hanami
         'mice'     => 'mouse',
         'oxen'     => 'ox',
         'people'   => 'person',
-        'quizzes'  => 'quiz',
         'sexes'    => 'sex',
         'testes'   => 'testis',
         'teeth'    => 'tooth',
@@ -385,6 +391,8 @@ module Hanami
           $1 + ES
         when /\A(.*)ss\z/
           $1 + SSES
+        when /\A(.*)#{Z}\z/
+          $1 + ZZES
         when /s\z/
           string
         else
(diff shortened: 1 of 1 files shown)


AlfonsoUceda:
I haven't added a test because `quizzes/quiz` is already tested, the problem was we didn't have a regular expression for that rule.

TiteiKo:
Humm won't this break things for words like "topaz => topazes" which will be pluralized as "topazzes", or "buzz => buzzes" which will be pluralized as "buzzzes"?

From what I can see here: http://www.enchantedlearning.com/grammar/partsofspeech/nouns/plurals/ second row of the table, quiz is an exception to the rule as it takes a second Z when plural, but other words ending with Z don't need the added Z...

So we have 2 problems here:

First, words ending in Z aren't correctly pluralized/singularized, example:

```
# resources :topazes, only: [:index, :show], should be /topazes and /topaz

             topazes GET, HEAD  /topazes                       Web::Controllers::Topazes::Index
              topaze GET, HEAD  /topazes/:id                   Web::Controllers::Topazes::Show
```

And second, the problem of prefixing irregular words, which is more general than with "quiz". See other examples below:

```
→ be hanami routes
                Name Method     Path                           Action

             quizzes GET, HEAD  /quizzes                       Web::Controllers::Quizzes::Index
                quiz GET, HEAD  /quizzes/:id                   Web::Controllers::Quizzes::Show
      prefix_quizzes GET, HEAD  /prefix_quizzes                Web::Controllers::PrefixQuizzes::Index
       prefix_quizze GET, HEAD  /prefix_quizzes/:id            Web::Controllers::PrefixQuizzes::Show
               sexes GET, HEAD  /sexes                         Web::Controllers::Sexes::Index
                 sex GET, HEAD  /sexes/:id                     Web::Controllers::Sexes::Show
          user_sices GET, HEAD  /user_sexes                    Web::Controllers::UserSexes::Index
            user_sex GET, HEAD  /user_sexes/:id                Web::Controllers::UserSexes::Show
              police GET, HEAD  /police                        Web::Controllers::Police::Index
              police GET, HEAD  /police/:id                    Web::Controllers::Police::Show
       state_polices GET, HEAD  /state_police                  Web::Controllers::StatePolice::Index
        state_police GET, HEAD  /state_police/:id              Web::Controllers::StatePolice::Show
```

AlfonsoUceda:
@TiteiKo you're right, grammar rules are difficult, so I think the best way is adding that exception in your code, the inflector covers the basics.

AlfonsoUceda:
Closing it!

TiteiKo:
Yep. But the inflector should still handle words ending with Z, only the rule should be Z => ZES, not Z => ZZES

AlfonsoUceda:
@TiteiKo I'll take a look later with that page in mind ;) thanks!

PR #168, opened 2016-11-09
Allow dasherized routes

childss:
When working with an API project following [JSON:API conventions](http://jsonapi.org/recommendations/), the common word separator is `-` (names and route segments are dasherized).

I'm very new to the Hanami codebase but it seems that routes are resolved to a controller in part via `Hanami::Utils::String#classify`. Currently that method treats any delimiter other than `_` as a namespace separator so `'test-resource'` becomes `Test::Resource` rather than `TestResource` which is awkward for naming and directory structure.

This change allows dasherized routes e.g. "test-resource" to behave like "test_resource".

Before this change dasherized routes resolve like:
```
'test-resource' => Web::Controllers::Test::Resource
'test_resource' => Web::Controllers::TestResource
```

After this change dasherized routes resolve like underscored routes:
```
'test-resource' => Web::Controllers::TestResource
'test_resource' => Web::Controllers::TestResource
```

In my testing so far this doesn't seem to cause any problems with my existing API app. I don't know if there is a larger Hanami integration test suite I should run this change against.

### DIFF
--- a/lib/hanami/utils/string.rb
+++ b/lib/hanami/utils/string.rb
@@ -72,6 +72,12 @@ module Hanami
       # @api private
       CLASSIFY_WORD_SEPARATOR = /#{CLASSIFY_SEPARATOR}|#{NAMESPACE_SEPARATOR}|#{UNDERSCORE_SEPARATOR}|#{DASHERIZE_SEPARATOR}/
 
+      # Regexp for #classify
+      #
+      # @since 0.9.0
+      # @api private
+      CLASSIFY_DELIMITER_REGEXP = /#{CLASSIFY_SEPARATOR}|#{DASHERIZE_SEPARATOR}/
+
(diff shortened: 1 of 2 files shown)


cllns:
Aesthetically, I like dash routes more than underscores. 

The only trouble is that this isn't the Ruby convention (in Rubygems specifically, but also elsewhere): http://guides.rubygems.org/name-your-gem/ 
I think we'd need a very compelling reason to break with the convention. Is there someway around this, perhaps an option in `hanami/router` instead?

childss:
Ah, yes I see the problem and I hadn't thought of that. Even if it doesn't directly break something in Hanami, it probably isn't a great idea to change this convention with `#classify`. I will go ahead and close this PR and investigate another option (perhaps a router option like you suggested).

Thanks!

PR #170, opened 2016-11-17
Introduced `Hanami::Utils::Json.parse` for safe JSON parsing.

### LINKED ISSUE #169: Use of JSON.load vs JSON.parse
The ruby docs explicitly say the following
> BEWARE: This method is meant to serialise data from trusted user input, like from your own database server or clients under your control, it could be **dangerous** to allow untrusted users to pass JSON sources into it.

Yet, `load` is used [here](https://github.com/hanami/utils/blob/master/lib/hanami/utils/json.rb), and hanami-router calls that method for parsing json bodies of http requests.

jodosha:
Introduced `Hanami::Utils::Json.parse` for safe JSON parsing.
Deprecate `Hanami::Utils::Json.load` in favor of `.parse`.

Introduced `Hanami::Utils::Json.generate` for safer JSON generation.
Deprecate `Hanami::Utils::Json.dump` in favor of `.generate`.

---

Closes #169 

/cc @AlfonsoUceda @beauby

### DIFF
--- a/lib/hanami/utils/json.rb
+++ b/lib/hanami/utils/json.rb
@@ -4,6 +4,8 @@ rescue LoadError
   require 'json'
 end
 
+require 'hanami/utils/deprecation'
+
 module Hanami
   module Utils
     # JSON wrapper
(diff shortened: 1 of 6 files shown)


lib/hanami/utils/json.rb
```diff
@@ -32,10 +58,26 @@ module Json
       # @raise [Hanami::Utils::Json::ParserError] if the paylod is invalid
       #
       # @since 0.8.0
+      #
+      # @deprecated Use {#parse} instead
       def self.load(payload)
+        Hanami::Utils::Deprecation.new("`Hanami::Utils::Json.load' is deprecated, please use `Hanami::Utils::Json.parse'")
```
AlfonsoUceda: do you think we need to add deprecate if we are under 1.0.0?

AlfonsoUceda:
@jodosha looks good :+1:

beauby:
If we move from `parse` to `load`, we might as well move from `dump` to `generate`, as `dump` is intended for marshalling as well.

lib/hanami/utils/json.rb
jodosha: @AlfonsoUceda I want to get this security patch out ASAP. That means to bump to a patch version: from 0.9.0 to 0.9.0.

Until now we followed the rule to introduce breaking changes only on minor releases: eg from 0.9.0 to 0.10.0, but this doesn't justify that.

But that would require to update the Upgrade notes on the site and we publish those only when we bump Hanami version, which isn't the case.

---

That's the solution to release it as 0.9.1, without any breaking change, hence the deprecation warning. When we'll release 1.0.0.beta1 we can remove `.load`.

jodosha:
@beauby Good point.

Issue #173, opened 2016-12-04
Pluralization problems for words ending in -en


katafrakt:
I'm using `Kitten` as entity in some of my tests and I was quite surprised when after updating to Hanami 0.9 I found `KittenRepository` complaining that table `kittina` was not present in my database.

I did some digging and found that [this rule](https://github.com/hanami/utils/blob/master/lib/hanami/utils/inflector.rb#L376-L377) is responsible for that. It changes `-en` ending into `-ina` in plural form. I had to do even more digging before I found a hint why it was introduced in the first place. The reason seems to be [some medical terms](http://www.meditec.com/resourcestools/medical-words/medical-plurals/) like `foramen` being `foramina` in plural. However, there are plain English words that are inflected regularly, such as aforementioned `kitten`, `mitten`, `omen` or `abdomen` (actually, `abdomina` is also correct plural form; on the other hand, so is [`foramens`](https://en.wiktionary.org/wiki/foramen#Noun)).

If I were to write a medical app, I would expect the need to set up custom pluralization rules with such words, however I was really surprised when I found that my `Kitten` is not inflected correctly. I would suggest to remove this rule, as it might be confusing for other people too.

cllns:
Thanks for this issue. I agree we should optimize for 🐱  and not `foramen` :) 

Feel free to open a PR for this if you have the time!

jodosha:
Fixed by #174

PR #174, opened 2016-12-04
Pluralize -en to -ens instead of -ina

katafrakt:
Plural rules with adding -s instead of changing ending to -ina are considered to be correct (i.e. "stamens" instead of "stamina") and do not create false plural forms of common English words, such as "kitten". They are still correctly singularized, though.

See: https://github.com/hanami/utils/issues/173

### DIFF
--- a/lib/hanami/utils/inflector.rb
+++ b/lib/hanami/utils/inflector.rb
@@ -93,10 +93,6 @@ module Hanami
 
       # @since 0.4.1
       # @api private
-      INA  = 'ina'.freeze
-
-      # @since 0.4.1
-      # @api private
       IS   = 'is'.freeze
 
       # @since 0.4.1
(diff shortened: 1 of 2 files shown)


katafrakt:
Travis fails because of new Rubocop version, introducing new checks. I may fix that, but I think it should belong in separate PR (as it may need a discussion about some new cops).

cllns:
👍 this looks good to me. 

This can be merged even though Travis is failing, because as Paweł said, it's just new Rubocop stuff.

AlfonsoUceda:
it makes sense this change, I prefer add custom rules for "weird" words than for normal words.

I agree with @cllns I think this can be merged and fix rubocop issues in another PR.

cc @jodosha

AlfonsoUceda:
@katafrakt could you rebase master in your branch? thanks!

katafrakt:
@AlfonsoUceda, rebased

jodosha:
@katafrakt Thank you for this fix. Merging now. 😄

PR #175, opened 2016-12-09
Add Gitter webhook to Travis config

cllns:
So we can get updates in Gitter when builds fail.

cc @AlfonsoUceda


davydovanton:
awesome 👍

PR #176, opened 2016-12-09
Fix (rubocop) build

cllns:
Don't require OrderedGems or EmptyMethods to be on same line (new rules introduced in latest version of Rubocop).


AlfonsoUceda:
thanks @cllns!

PR #179, opened 2016-12-13
Deprecate Hanami::Interactor::Result#failing?

### LINKED ISSUE #172: Interactor uses gramatically wrong terms for success/failure
Thank you for your work on this project!

When an Interactor call was a success, we check with either

```ruby
result = Signup.new().call
result.success?
# OR
result.successful?
```

but we check failure with

```ruby
result.failing?
```

This doesn't make grammatic sense, because the opposite of `success?` would be `failure?` and the opposite of `successful?` would still be `failure?`.  However, `failing?` is definitely not correct, because its opposite would be `succeeding?`.

I think this is quite confusing, even though it is a small nitpick.  Maybe `failing?` could be deprecated and aliased to `failure?`?

Thoughts? Thanks again!

mgrachev:
Closes #172

### DIFF
--- a/lib/hanami/interactor.rb
+++ b/lib/hanami/interactor.rb
@@ -1,6 +1,7 @@
 require 'hanami/utils/basic_object'
 require 'hanami/utils/class_attribute'
 require 'hanami/utils/hash'
+require 'hanami/utils/deprecation'
 
 module Hanami
   # Hanami Interactor
@@ -20,6 +21,7 @@ module Hanami
       METHODS = ::Hash[initialize:  true,
                        success?:    true,
                        successful?: true,
+                       failure?:    true,
                        failing?:    true,
                        fail!:       true,
                        prepare!:    true,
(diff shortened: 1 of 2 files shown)


lib/hanami/interactor.rb
```diff
@@ -57,7 +59,19 @@ def successful?
...
       def failing?
+        Utils::Deprecation.new("`Hanami::Interactor::Result#failing?' is deprecated, please use `Hanami::Interactor::Result#failure?'")
+        failure?
+      end
+
+      # Check if the current status is not successful
+      #
+      # @return [TrueClass,FalseClass] the result of the check
+      #
+      # @since 0.9.2
```
AlfonsoUceda: @mgrachev  could you put to x.x.x?

TiteiKo:
Looks good to me 👍

lib/hanami/interactor.rb
mgrachev: Done.

cllns:
Thanks @mgrachev!

PR #181, opened 2017-01-05
Improve logger

AlfonsoUceda:
Talking with @jodosha I saw this example log for default formatter in hanami logger:
```
app=hanami.model severity=INFO time=2017-01-03 17:17:17 UTC message=(0.000260s) SELECT `id`, `name`, `created_at`, `updated_at` FROM `authors` WHERE (`id` IN (1)) ORDER BY `authors`.`id`
```
This kind of log, will break with integrations like logstash, if your value has spaces, then it should be wrapped between quotes. (I'm using single quotes).

For time, I'm using RFC 3339 in UTC without spaces.

The result of the previous log would be:
```
app=hanami.model severity=INFO time=2017-01-03T17:17:17+00:00 message='(0.000260s) SELECT `id`, `name`, `created_at`, `updated_at` FROM `authors` WHERE (`id` IN (1)) ORDER BY `authors`.`id`'
```

cc @hanami/contributors @hanami/core

### DIFF
--- a/lib/hanami/logger.rb
+++ b/lib/hanami/logger.rb
@@ -108,10 +108,6 @@ module Hanami
       # @api private
       NEW_LINE = $/
 
-      # @since x.x.x
-      # @api private
-      TIME_RFC3339_FORMAT = "%FT%T%:z".freeze
-
       include Utils::ClassAttribute
 
       class_attribute :subclasses
@@ -149,7 +145,7 @@ module Hanami
         _format({
           app:      @application_name,
           severity: severity,
-          time:     time.utc.strftime(TIME_RFC3339_FORMAT)
+          time:     time
         }.merge(
           _message_hash(msg)
         ))
@@ -202,6 +198,7 @@ module Hanami
       # @since 0.8.0
       # @api private
       def _format(hash)
+        hash[:time] = hash[:time].utc.iso8601
         Hanami::Utils::Json.dump(hash)
       end
     end
@@ -211,7 +208,7 @@ module Hanami
     #
     # @since 0.5.0
     # @api private
-    DEFAULT_APPLICATION_NAME = 'Hanami'.freeze
+    DEFAULT_APPLICATION_NAME = 'hanami'.freeze
 
     # @since 0.8.0
     # @api private
--- a/lib/hanami/logger.rb
+++ b/lib/hanami/logger.rb
@@ -159,10 +159,8 @@ module Hanami
 
       # @since 0.8.0
       # @api private
-      def _message_hash(message) # rubocop:disable Metrics/MethodLength
+      def _message_hash(message)
         case message
-        when Hash
-          message
         when Exception
           Hash[
             message:   message.message,
@@ -174,10 +172,18 @@ module Hanami
         end
       end
 
-      # @since 0.8.0
+      # @since x.x.x
       # @api private
       def _format(hash)
-        hash.map { |k, v| "#{k}=#{v}" }.join(SEPARATOR) + NEW_LINE
+        format = "[#{hash[:app]}] [#{hash[:severity]}] [#{hash[:time]}]"
+        format << " #{hash[:error]}:" if hash.key?(:error)
+        format << " #{hash[:message]}#{NEW_LINE}"
+        if hash.key?(:backtrace)
+          hash[:backtrace].each do |line|
+            format << "from #{line}#{NEW_LINE}"
+          end
+        end
+        format
       end
     end
 
--- a/lib/hanami/logger.rb
+++ b/lib/hanami/logger.rb
@@ -108,6 +108,10 @@ module Hanami
       # @api private
       NEW_LINE = $/
 
+      # @since x.x.x
+      # @api private
+      TIME_RFC3339_FORMAT = "%FT%T%:z".freeze
+
       include Utils::ClassAttribute
 
       class_attribute :subclasses
@@ -145,7 +149,7 @@ module Hanami
         _format({
           app:      @application_name,
           severity: severity,
-          time:     time.utc
+          time:     time.utc.strftime(TIME_RFC3339_FORMAT)
         }.merge(
           _message_hash(msg)
         ))
--- a/lib/hanami/logger.rb
+++ b/lib/hanami/logger.rb
@@ -168,7 +168,7 @@ module Hanami
         end
       end
 
-      # @since x.x.x
+      # @since 0.8.0
       # @api private
       def _format(hash)
         format = "[#{hash[:app]}] [#{hash[:severity]}] [#{hash[:time]}]"
--- a/test/logger_test.rb
+++ b/test/logger_test.rb
@@ -274,7 +274,7 @@ describe Hanami::Logger do
 
     it 'has default app tag when not in any namespace' do
       class TestLogger < Hanami::Logger; end
-      TestLogger.new.application_name.must_equal 'Hanami'
+      TestLogger.new.application_name.must_equal 'hanami'
     end
 
     it 'infers apptag from namespace' do
@@ -312,7 +312,7 @@ describe Hanami::Logger do
               class TestLogger < Hanami::Logger; end
               TestLogger.new(formatter: nil).info('foo')
             end
-          output.must_equal "[Hanami] [INFO] [1988-09-01T00:00:00+00:00] foo\n"
+          output.must_equal "[hanami] [INFO] [2017-01-15 16:00:23 +0100] foo\n"
         end
       end
     end
@@ -328,7 +328,7 @@ describe Hanami::Logger do
                 class TestLogger < Hanami::Logger; end
                 TestLogger.new(formatter: :json).info('foo')
               end
-            output.must_equal '{"app":"Hanami","severity":"INFO","time":"1988-09-01T00:00:00+00:00","message":"foo"}'
+            output.must_equal '{"app":"hanami","severity":"INFO","time":"2017-01-15T15:00:23Z","message":"foo"}'
           end
         end
       end
@@ -343,7 +343,7 @@ describe Hanami::Logger do
                 class TestLogger < Hanami::Logger; end
                 TestLogger.new(formatter: Hanami::Logger::JSONFormatter.new).info('foo')
               end
-            output.must_equal '{"app":"Hanami","severity":"INFO","time":"1988-09-01T00:00:00+00:00","message":"foo"}'
+            output.must_equal '{"app":"hanami","severity":"INFO","time":"2017-01-15T15:00:23Z","message":"foo"}'
           end
         end
       end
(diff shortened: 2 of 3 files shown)


davydovanton:
looks good for me 👍

jodosha:
@AlfonsoUceda Thanks for this PR. Where's the RFC 3339 comes from? Is that a de facto standard? For JSON it's ISO 8601.

AlfonsoUceda:
You can see the RFC here https://tools.ietf.org/html/rfc3339

You have more info here: http://stackoverflow.com/questions/522251/whats-the-difference-between-iso-8601-and-rfc-3339-date-formats

lib/hanami/logger.rb
```diff
@@ -145,7 +145,7 @@ def call(severity, time, _progname, msg)
         _format({
           app:      @application_name,
           severity: severity,
-          time:     time.utc
+          time:     time.utc.to_datetime.rfc3339
```
jodosha: The problem here is the creation of intermediate objects:

```ruby
irb(main):003:0> now = Time.now
=> 2017-01-05 18:05:09 +0100
irb(main):004:0> now.object_id
=> 70259928409580
irb(main):005:0> now.utc.object_id
=> 70259928409580
irb(main):006:0> now.utc.object_id
=> 70259928409580
irb(main):007:0> now.utc.to_datetime.object_id
=> 70259928246620
irb(main):008:0> now.utc.to_datetime.object_id
=> 70259928186200
irb(main):009:0> now.utc.to_datetime.rfc3339.object_id
=> 70259928152180
irb(main):010:0> now.utc.to_datetime.rfc3339.object_id
=> 70259928064880
```

As you can see, when `time` is converted in UTC, it's still keeping the same Object ID, which is good. But for `to_datetime` and `rfc3339`, it creates two new objects.

If we want to keep RFC 3339, we should avoid to create the intermediary `DateTime` object.

test/logger_test.rb
```diff
@@ -446,7 +446,18 @@ class TestLogger < Hanami::Logger; end
...
+          output.must_equal "app=Hanami severity=INFO time=1988-09-01T00:00:00+00:00 message=\'[\"foo\"]\'\n"
+        end
+      end
+
+      it 'wraps messages with spaces into single quotes' do
+        stub_time_now do
+          output =
+            stub_stdout_constant do
+              class TestLogger < Hanami::Logger; end
+              TestLogger.new.info('This is a message with spaces')
```
jodosha: @AlfonsoUceda What happens when a message with `'` is passed? We should escape it, right now it returns:

```ruby
TestLogger.new.info("This isn't a message without '")
```

```
"app=Hanami severity=INFO time=1988-09-01T00:00:00+00:00 message='This isn't a message without ''
 "
```

AlfonsoUceda: yeah, I was wondering that in my mind, I'll take a look

AlfonsoUceda:
@jodosha I used a strftime for the time and now the message is quoted. Please, check when you can.

jodosha:
@AlfonsoUceda My question was more, why that specific format is used for logging?

test/logger_test.rb
```diff
@@ -455,9 +455,9 @@ class TestLogger < Hanami::Logger; end
           output =
             stub_stdout_constant do
               class TestLogger < Hanami::Logger; end
-              TestLogger.new.info('This is a message with spaces')
+              TestLogger.new.info("This is a message with quotes \" ' and with spaces")
```
jodosha: This is cheating :wink: 

Shall we try with `TestLogger.new.info(%(This is a message with quotes " ' and with spaces))` instead? I'm afraid that we're going down the rabbit hole. We should find a way to escape strings.

AlfonsoUceda:
@jodosha I don't know, I'll ask my devops mates if they know something ;)

test/logger_test.rb
AlfonsoUceda: @jodosha with that the test passes, of course we can make a gsub to scape inner single quotes if we wrap with them or viceversa.

AlfonsoUceda:
@jodosha searching in syslog protocol I found this http://www.rfc-editor.org/rfc/rfc5424.txt section 
> 6.2.3.  TIMESTAMP

I extracted that from http://www.employees.org/~lonvick/index.shtml

test/logger_test.rb
jodosha: @AlfonsoUceda I was joking :wink: 

Of course it passes, because I used an alternative Ruby syntax.
My point is about the output that is generated by that escaping.

Let me try to explain with code. If in the code you add the following line:

```ruby
File.open("log", "w+") { |f| f.write(output) }
```

Then if you look at `log`, here's the result:

```
app=Hanami severity=INFO time=1988-09-01T00:00:00+00:00 message="This is a message with quotes " ' and with spaces"
```

If you look at the message, is interrupted: the first `"` matches with the one in the middle of the sentence. So the message is: `This is a message with quotes `, while the remaining part `' and with spaces"` would get a parser crazy.

AlfonsoUceda: @jodosha I've updated the PR, I use gsub to replace '"' with '\"', but the only problem I see I have to check the message with `is_a?`, but it works now.

AlfonsoUceda: @jodosha I was wondering about this test, why put an array in the message? I understand a hash, because it could be parsed as json but Array?

jodosha:
@AlfonsoUceda Thanks for reiterating this PR.

The problem is a whack-a-mole: when we solve a problem, the next one arises.

Before to go into the weeds with complex regular expressions to escape all the possible inputs, and then maintain it, we should stop now and answer the following questions.

---

#### What's the current situation?

As now, we have two formatters: `:default` and `:json`.

The first one has the purpose of being human readable, useful for development purposes.
The latter is designed to me "machine oriented", and therefore parseable.

A developer can create their own formatter and use it.

```ruby
module Web
  class Application < Hanami::Application
    configure do
      logger.format MyCustomFormatter.new
    end
  end
end
```

#### Why we are introducing this change?

Because we want to make `:default` formatter parseable.

But IMO, our choice for parseable format is and should remain `:json`. Because it's structured and we can delegate to Ruby's `JSON` the responsibility of escaping quotes.

#### Is the current implementation of `:default` formatter human readable enough?

Yes.

But it can be DRY'd, by removing the keys (se below) and maybe adding colours when the output stream is a TTY (like `$stdout`).

An example could be:

From:

```
app=hanami.model severity=INFO time=2017-01-03 17:17:17 UTC message=(0.000260s) SELECT `id`, `name`, `created_at`, `updated_at` FROM `authors` WHERE (`id` IN (1)) ORDER BY `authors`.`id`
```

To:

```
[hanami.model] [INFO] [2017-01-03T17:17:17Z] (0.000260s) SELECT `id`, `name`, `created_at`, `updated_at` FROM `authors` WHERE (`id` IN (1)) ORDER BY `authors`.`id`
```

PS: imagine this with colours in a shell.

---

**TL;DR:** My opinion is to **not** wrap log messages in quotes, but to change the `:default` format to something **more** human readable and eye-catching. Parseable logger format should remain `:json` for production. Other cases can be handled with a custom logger format.

@AlfonsoUceda @TiteiKo What's your opinion?

AlfonsoUceda:
@jodosha it makes sense :+1: so what do you think if we only apply the timestamp change?

joneslee85:
@AlfonsoUceda my 5cent on this is I think hanami should only support JSON formatter. It would argue that JSON is human readable, parsing it to `jq` would prettify the STDOUT.

Having said that, if developers want to format it in their own way, they could always write logger middleware that takes in JSON input.

screen shot 2017-01-11 at 12 34 30 pm

AlfonsoUceda:
thanks @joneslee85 for your input, yeah, I think it's the best way, use JSON as default in production ;)

jodosha:
@joneslee85 That is an output that goes directly to developer's console. We want the default to be human readable without any extra tool. Does `jq` supports pipes like `tail -f`?

jodosha:
@AlfonsoUceda 
> @jodosha it makes sense 👍 so what do you think if we only apply the timestamp change?

Can you also please apply all the other changes to the default logger? I mean, the format that I suggested above.

Please don't forget that the arguments of `#strftime` should be extracted into a constant.

lib/hanami/logger.rb
```diff
@@ -108,6 +108,10 @@ class Formatter < ::Logger::Formatter
       # @api private
       NEW_LINE = $/
 
+      # @since x.x.x
+      # @api private
+      TIME_RFC3339_FORMAT = "%FT%T%:z".freeze
```
jodosha: ❤️

lib/hanami/logger.rb
```diff
@@ -174,10 +172,18 @@ def _message_hash(message) # rubocop:disable Metrics/MethodLength
         end
       end
 
-      # @since 0.8.0
+      # @since x.x.x
```
jodosha: @AlfonsoUceda I'm gonna cleanup this. But remember, if a method was already existing and it only changes the implementation, we shouldn't retag with `@since x.x.x`, but keeping it the old version. 😉

lib/hanami/logger.rb
```diff
@@ -145,7 +149,7 @@ def call(severity, time, _progname, msg)
         _format({
           app:      @application_name,
           severity: severity,
-          time:     time.utc
+          time:     time.utc.strftime(TIME_RFC3339_FORMAT)
```
jodosha: @AlfonsoUceda I think we should pass just `time` to the formatters and let it to decide which timezone and format to use.

---

**Default formatter:** We said that default formatter is designed for human readability during the development phase.

While UTC is a timezone that it's a standard in production, it's not human friendly if you live far away from it and you're developing your project.

RFC 3339 is a standard for logging parsers, but we want to give up with parsing for this formatter.

IMO the default Ruby formatting for `Time` is enough, and it safes a few CPU cycles as we don't need to use `#strftime`.

```ruby
irb(main):001:0> now = Time.now
=> 2017-01-14 16:14:56 +010 # More readable, no CPU waste
irb(main):002:0> now.strftime("%FT%T%:z")
=> 2017-01-14T16:14:56+01:00 # Less readable, more CPU cycles
```

---

**JSON formatter:** This is for production (convert to UTC), and it must be parseable. In this case it makes sense to format the timestamp. I'm only wondering if in this case we should use ISO 8601, which is the de-facto standard format for JSON.

---

**Custom formatter:** The win of passing the `time` object as it is, is to give the freedom to developers to leave the current timezone or convert it to UTC. Also, they can decide their own time format.

---

WDYT?

lib/hanami/logger.rb
AlfonsoUceda: @jodosha I didn't remember that sorry :P

lib/hanami/logger.rb
AlfonsoUceda: 👍

AlfonsoUceda:
@jodosha as we talked in Gitter, please go ahead with those changes, I'm going to focus in router. Thanks!

AlfonsoUceda:
@jodosha :+1:

PR #184, opened 2017-01-08
Remove deprecation warning for logger json format output

bhanuone:
update deprecated method `dump` to `generate`


bhanuone:
@jodosha

jodosha:
@bhanuone Thanks for this PR! I'm merging it.

Ref #170

PR #185, opened 2017-01-09
Copyedit doc for Hanami::Utils::BasicObject#pretty_print

b264:
as seen at  http://www.rubydoc.info/gems/hanami-utils/Hanami/Utils/BasicObject#pretty_print-instance_method


jodosha:
@b264 Thank you! 👍

PR #191, opened 2017-02-20
Introduce ::filled? method

davydovanton:
I think it will be helpful to have the antonym for `.blank?` method instead of using `!Hanami::Utils::Blank.blank?(value)`.

WDYT?

/cc @hanami/core @hanami/ecosystem

### DIFF
--- a/test/blank_test.rb
+++ b/test/blank_test.rb
@@ -23,4 +23,22 @@ describe Hanami::Utils::Blank do
        end
      end
   end
+
+  describe '.filled?' do
+    [nil, false, '', '   ', "  \n\t  \r ", '　', "\u00a0", [], {}, Set.new,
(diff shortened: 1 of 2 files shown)


test/blank_test.rb
```diff
@@ -23,4 +23,22 @@
        end
      end
   end
+
+  describe '.filled?' do
+    [nil, false, '', '   ', "  \n\t  \r ", '　', "\u00a0", [], {}, Set.new,
+     Hanami::Utils::Kernel.Boolean(0), Hanami::Utils::String.new(''),
+     Hanami::Utils::Hash.new({})].each do |v|
+       it 'returns true' do
```
TiteiKo: Shouldn't it be `it 'returns false'` ? (and the opposite in the next test)

davydovanton:
@jodosha done. Thanks for review 🎉

PR #193, opened 2017-03-08
Remove Hanami::Utils::Attributes as no longer used

jodosha:


davydovanton:
@jodosha where we used this class before?

AlfonsoUceda:
@davydovanton If I remember correctly, in validations

test/class_attribute_test.rb
```diff
@@ -1,152 +0,0 @@
-require 'test_helper'
-require 'hanami/utils/class_attribute'
-
-describe Hanami::Utils::ClassAttribute do
```
davydovanton: I think you need to delete `Attributes` tests too :)

jodosha: OMG, this was the wrong file to remove 😱

PR #199, opened 2017-03-21
Interactor Result can check for failure reason

juandazapata:
This commit enables a more expressive way to interact with a failure
reason.

When the interactor is failed with the `error!` method, we can check in
the result.

Inside the interactor, we fail with `error!` and the error reason.
```ruby
def call
  error!('invalid_credit_card') if payment_gateway.invalid_card?
  error!('expired_credit_card') if whatever?
  # etc...
end
```

Then when calling the interactor, we can check for the failure with a given error
```ruby
# action
def a_controller_action
  result = MyInteractor.new(some_params).call

  render :new     if result.failure?('invalid_credit_card')
  render :expired if result.failure?('expired_credit_card')
  render :success
end
```

### DIFF
--- a/lib/hanami/interactor.rb
+++ b/lib/hanami/interactor.rb
@@ -55,11 +55,14 @@ module Hanami
 
       # Check if the current status is not successful
       #
+      # @param reason [Object] the reason of the failure to check
+      #
       # @return [TrueClass,FalseClass] the result of the check
       #
       # @since 0.9.2
-      def failure?
-        !successful?
+      def failure?(reason = nil)
+        return !successful? if reason.nil?
+        !successful? && (error == reason)
       end
 
       # Force the status to be a failure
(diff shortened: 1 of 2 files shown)


TiteiKo:
Hi there,

You can already achieve something similar:

```ruby
def call
  error!('invalid_credit_card') if payment_gateway.invalid_card?
  error!('expired_credit_card') if whatever?
end
```

```ruby
def a_controller_action
  result = MyInteractor.new(some_params).call

  render :new     if result.error == 'invalid_credit_card'
  render :expired if result.error == 'expired_credit_card'
  render :success
end
```

juandazapata:
Hi @TiteiKo , thanks for reviewing this. Yes, I'm aware of that option. This PR aims to add a little more expressiveness to the failure check, but it's essentially the same functionality.

jodosha:
@juandazapata Thank you for this PR, but I'm not merging it.

We're now in feature freeze before the 1.0, so we won't change the API in this phase.

Also, we need more solid error checking, with proper success/failure objects instead of checking with strings.

I hope you can understand. Thank you again.

juandazapata:
Fair enough @jodosha , thanks for reviewing and thanks for creating hanami!

PR #200, opened 2017-03-24
Migrate from Minitest to RSpec

artofhuman:
/cc @jodosha 

I can pass this test https://github.com/hanami/utils/blob/develop/test/isolation/json/json_test.rb#L14 in `rspec`. I think i need suggestion with it. In this case, `multijson` should not be load, but it loads.


.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: ruby
 sudo: false
 cache: bundler
-script: 'script/ci && bundle exec rubocop'
+script: 'bundle exec rake spec:coverage --trace && bundle exec rubocop'
```
jodosha: This should stay `script/ci`. It guarantees isolation for `multi_json` specs et al.

jodosha:
@artofhuman Thanks for this PR, it fails because you removed `script/ci` from `.travis.yml`.

artofhuman:
Now ci show the broken test, also rubocop updated to 0.48 and broke build too.

jodosha:
@artofhuman Anything I can help with?

jodosha:
@artofhuman RE isolation tests. I had the same problem with `hanami-controller`. Here's how I fixed it: https://github.com/hanami/controller/commit/7d445e0d35333db7d87ed9cedb2b7a6f61d29eb3

artofhuman:
@jodosha thanks for the link. I think it's ready for review.

PR #203, opened 2017-03-27
Remove references to removed `failing?` method

cllns:
It's now just `failure?`.

Clean up for #194


jodosha:
The build failed because of Rubocop, I'm merging this and right after #204, which fixes these failures.

PR #205, opened 2017-03-30
Logger arguments

### LINKED ISSUE #201: Add logrotation to Hanami::Logger
I propose to add logrotation to Hanami::Logger to support limiting the logs in duration or size. Same as standard [Ruby logger](https://ruby-doc.org/stdlib-2.3.0/libdoc/logger/rdoc/Logger.html#class-Logger-label-How+to+create+a+logger).

jodosha:
Allow to pass arbitrary arguments to the logger, so they can be forwarded to the superclass.

### Periodic rotation

```ruby
require 'hanami/logger'

# Rotate daily
logger = Hanami::Logger.new('bookshelf', 'daily', stream: 'development.log')
```

### File size rotation

```ruby
require 'hanami/logger'

# leave 10 old log files where the size is about 1,024,000 bytes
logger = Hanami::Logger.new('bookshelf', 10, 1024000, stream: 'development.log')
```

See https://ruby-doc.org/stdlib/libdoc/logger/rdoc/Logger.html#class-Logger-label-How+to+create+a+logger

---

Closes #201 /cc @sebastjan-hribar and @hanami/core for review.

### DIFF
--- a/lib/hanami/logger.rb
+++ b/lib/hanami/logger.rb
@@ -256,10 +256,125 @@ module Hanami
     # @param application_name [String] an optional application name used for
     #   tagging purposes
     #
-    # @param stream [String, IO, StringIO, Pathname] an optional log stream. This is a filename
-    # (String) or IO object (typically `$stdout`, `$stderr`, or an open file).
+    # @param *args [Array<Object>] an optional set of arguments to honor Ruby's
+    #   `Logger#initialize` arguments. See Ruby documentation for details.
+    #
+    # @param stream [String, IO, StringIO, Pathname] an optional log stream.
+    #   This is a filename (`String`) or `IO` object (typically `$stdout`,
+    #   `$stderr`, or an open file). It defaults to `$stderr`.
+    #
+    # @param level [Integer,String] logging level. It can be expressed as an
+    #   integer, according to Ruby's `Logger` from standard library or as a
+    #   string with the name of the level
+    #
+    # @param formatter [Symbol,#_format] a formatter - We support `:json` as
+    #   JSON formatter or an object that respond to `#_format(data)`
     #
     # @since 0.5.0
+    #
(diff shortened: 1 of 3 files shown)


sebastjan-hribar:
Hi @jodosha 

this is perfect and exactly like I proposed. Thank you for your efforts!

lib/hanami/logger.rb
```diff
@@ -256,12 +256,127 @@ def _format(hash)
...
+    #
+    #   # => "[Hanami] [DEBUG] [2017-03-30 15:55:22 +0200] Hello World\n"
+    #
+    # @example JSON formatter
+    #   require 'hanami/logger'
+    #
+    #   logger = Hanami::Logger.new(formatter: :json)
+    #   logger.info "Hello World"
+    #
+    #   # => {"app":"Hanami","severity":"DEBUG","time":"2017-03-30T13:57:59Z","message":"Hello World"}
```
davydovanton: Kudos to new awesome logger documentations 👏

test/logger_test.rb
```diff
@@ -26,6 +27,26 @@ class TestLogger < Hanami::Logger; end
...
+        stream.dirname.mkpath
+      end
+
+      let(:stream) { Pathname.new(Dir.pwd).join('tmp', 'rotation.log') }
+      let(:count)  { 5 }
+
+      it 'passes them to the superclass' do
+        logger = Hanami::Logger.new('rotation', count, 512, stream: stream)
+
+        1_000.times do
```
davydovanton: WDYT if we reduce this value to `100` for improving test speed?

jodosha: Ok. Here's before and after numbers:

```
./script/ci  4.45s user 1.03s system 78% cpu 6.943 total
```

```
./script/ci  4.19s user 0.81s system 95% cpu 5.219 total
```

davydovanton: 1.7 seconds win! :D

PR #207, opened 2017-04-07
Pretty format form_params hash when using default Logger::Formatter

gizotti:
This PR is an extension to hanami/hanami#756 and update `Logger::Formatter`, used as default logger on `development` environment, to pretty format the the hash passed as `form_params`. 

```
[hanami_blog] [INFO] [2017-04-07 18:29:40 +1000] Parameters: {
  "post": {
    "title": "My Title",
    "body": "My Body",
    "password": "[FILTERED]",
    "password_confirmation": "[FILTERED]"
  }
}
```

My only question is if we should make we should format the values of the `form_params` key, or if we should set the logger as `pretty_formattable` and format any `Hash` that is passed?

### DIFF
--- a/lib/hanami/logger.rb
+++ b/lib/hanami/logger.rb
@@ -187,7 +187,7 @@ module Hanami
         return _format_error(result, hash) if hash.key?(:error)
 
         values = hash.each_with_object([]) do |(k, v), memo|
-          memo << v unless RESERVED_KEYS.include?(k)
+          memo << _formatted_log_value(k, v) unless RESERVED_KEYS.include?(k)
         end
 
         result << " #{values.join(SEPARATOR)}#{NEW_LINE}"
@@ -205,6 +205,15 @@ module Hanami
         end
 
         result
+      end
+
+      # @api private
+      def _formatted_log_value(k, v)
+        if k == :form_params
+          "Parameters: #{JSON.pretty_generate(v)}"
+        else
+          v
+        end
       end
     end
 
--- a/lib/hanami/logger.rb
+++ b/lib/hanami/logger.rb
@@ -4,6 +4,7 @@ require 'logger'
 require 'hanami/utils/string'
 require 'hanami/utils/json'
 require 'hanami/utils/class_attribute'
+require 'hanami/utils/blank'
 
 module Hanami
   # Hanami logger
@@ -92,6 +93,7 @@ module Hanami
   #   Hanami::Logger.new(formatter: Hanami::Logger::JSONFormatter).info('Hello')
   #   # => "{\"app\":\"Hanami\",\"severity\":\"INFO\",\"time\":\"1988-09-01 00:00:00 UTC\",\"message\":\"Hello\"}"
   class Logger < ::Logger
+    class InvalidFilteredParameterTypeException < StandardError; end
     # Hanami::Logger default formatter.
     # This formatter returns string in key=value format.
     #
@@ -99,7 +101,7 @@ module Hanami
     # @api private
     #
     # @see http://www.ruby-doc.org/stdlib/libdoc/logger/rdoc/Logger/Formatter.html
-    class Formatter < ::Logger::Formatter
+    class Formatter < ::Logger::Formatter # rubocop:disable Metrics/ClassLength
       # @since 0.8.0
       # @api private
       SEPARATOR = ' '.freeze
@@ -117,15 +119,20 @@ module Hanami
       class_attribute :subclasses
       self.subclasses = Set.new
 
-      def self.fabricate(formatter, application_name)
-        case formatter
-        when Symbol
-          (subclasses.find { |s| s.eligible?(formatter) } || self).new
-        when nil
-          new
-        else
-          formatter
-        end.tap { |f| f.application_name = application_name }
+      def self.fabricate(formatter, application_name, filter) # rubocop:disable Metrics/MethodLength
+        fabricated_formatter = case formatter
+                               when Symbol
+                                 (subclasses.find { |s| s.eligible?(formatter) } || self).new
+                               when nil
+                                 new
+                               else
+                                 formatter
+                               end
+
+        fabricated_formatter.application_name = application_name
+        fabricated_formatter.filter           = filter
+
+        fabricated_formatter
       end
 
       # @api private
@@ -146,6 +153,10 @@ module Hanami
       # @since 1.0.0
       # @api private
       attr_reader :application_name
+
+      # @since x.x.x
+      # @api private
+      attr_writer :filter
 
       # @since 0.5.0
       # @api private
@@ -168,7 +179,7 @@ module Hanami
       def _message_hash(message) # rubocop:disable Metrics/MethodLength
         case message
         when Hash
-          message
+          _filtered_message(message)
         when Exception
           Hash[
             message:   message.message,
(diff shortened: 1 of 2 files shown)


lib/hanami/logger.rb
```diff
@@ -187,7 +187,14 @@ def _format(hash)
         return _format_error(result, hash) if hash.key?(:error)
 
         values = hash.each_with_object([]) do |(k, v), memo|
-          memo << v unless RESERVED_KEYS.include?(k)
+          next if RESERVED_KEYS.include?(k)
+
+          memo << case k
```
davydovanton: maybe is better to use `if ... else ... end` here?

davydovanton: or create general method for this condition i.e:

```ruby
memo << formatting_method(k, v) unless RESERVED_KEYS.include?(k)

def formatting_method(k, v)
  if k == :form_params
    "Parameters: #{JSON.pretty_generate(v)}"
  else
    v
  end
end
```

gizotti: 👍 for the second option. Will go with it.

gizotti:
@davydovanton what's up with CI and rubocop? I haven't touched anything that is failing on CI.

davydovanton:
@gizotti if I understand right it's rubocop warnings
https://travis-ci.org/hanami/utils/jobs/219644941#L225

gizotti:
@davydovanton yes, but I am wondering why these are happening if I haven't changed them. Should I fix it with this PR?

davydovanton:
@gizotti good question 👍 
@hanami/core WDYT?

gizotti:
Figured it out. The problem here is that I created this branch from the `develop` branch, but rubocop updated to `0.48`.

All these rubocop warnings have been fixed in master, so should I just rebase my branch from the `master` branch?

gizotti:
@davydovanton A rebase from master sorted it out :) Should be good to go now.

gizotti:
My pleasure. I'm glad I can help 💚.

TiteiKo:
Humm I'm not sure that it's a good idea... Sure it's nicer on development environment, but this formatter is also used in production by some ppl and this is a breaking change that potentially won't be parsable at all by integrations tools like Logstash

ping @AlfonsoUceda that is directly concerned by this and will know best

AlfonsoUceda:
@Titeiko I was concerned at the beginning but we can change our formatter for production environments.

gizotti:
As requested by @jodosha I've moved the filtering responsibility to the `Logger` instead of the framework.

The `message` is filtered before being formatted in the `call` method of the base `Formatter` class. Any `Hash` value of the `message` `Hash` will be filtered if `filter` is passed on the `Logger` initialization. 
Not only the `form_hash` will be filtered, but any hash logged.

gizotti:
@jodosha no worries. My pleasure to help :)

`:form_params` also ensure that we add the `Parameters:` text to the logs. I like the idea of pretty print nested messages, but we lose the `Parameters:` text from it. Any suggestion about this?

---

`JSON.pretty_generate` on the default formatter was based on the idea of using the default formatter in `development` and the json formatter in `production`.
This comes back to @TiteiKo's concert:
> Humm I'm not sure that it's a good idea... Sure it's nicer on development environment, but this formatter is also used in production by some ppl and this is a breaking change that potentially won't be parsable at all by integrations tools like Logstash
ping @AlfonsoUceda that is directly concerned by this and will know best

Should we have another option in the `Formatter` to configure if it's pretty formatted or not, independent if it's the default or the json formater?

---

About the string messages. What if we get a message like the following:

`TestLogger.new(filters: "credit_card").info(%(credit_card=111122223334444, name=Gabriel))`

Should we filter the whole message?

---

I'll make all the changes once we come to a concrete decision about all of this :)

jodosha:
@gizotti Thanks for the quick reply

> :form_params also ensure that we add the Parameters: text to the logs. I like the idea of pretty print nested messages, but we lose the Parameters: text from it. Any suggestion about this?

The idea is to have no difference between route params and POST params. The router/action make them available inside one Hash: `params`. This object wraps a (nested) hash where all the params are hold without any difference from their source. Example: you can access transparently `params[:id]` (route param from `/books/:id`) and `params.get(:book, :title)` (payload param from `PATCH` request).

Long story short: we already have params in one _data bag_ (`params`), can we just pretty print and filter without having special rules like `:form_params`?

---

> JSON.pretty_generate on the default formatter was based on the idea of using the default formatter in development and the json formatter in production.

I'm confused about this statement. 😅  It looks to me that we're using JSON formatting for both default formatter and for the JSON one.

To me the expectation is to use JSON formatting only for JSON formatter..

---

> About the string messages. What if we get a message like the following:
> 
> TestLogger.new(filters: "credit_card").info(%(credit_card=111122223334444, name=Gabriel))
> 
> Should we filter the whole message?

I'm not 100% sure about this, let's skip for now.

---

Thank you very much.

TiteiKo:
There's two distinct things here:
* adding request body parameters to the log
* prettyfying the output

## Pretty print
@gizotti Regarding the pretty print part, we won't add it in the default formatter as it would either be breaking the log one-liner format or add too much complexity (option to pretty print or not). You can however easily create and use your own formatter:

```ruby
require 'hanami/logger'

class PrettyFormatter < Hanami::Logger::Formatter
  def self.eligible?(name)
    name == :pretty
  end

  def _format(*)
    "My Log Format\n"
  end
end

# in environment.rb
environment :development do
  logger level: :debug, formatter: :pretty
end
```

## Request body parameters

So, I've dug a bit in this issue, and it looks like [where we actually log the request](https://github.com/hanami/hanami/blob/aec069b602c772e279aa0a7f48d1a04d01756ee3/lib/hanami/common_logger.rb#L48), we don't have access to the action params. (and this is logical as the request could be a 404)

We have however access to the parameters seen by rack as a hash and they can easily be printed.

The result of my first draft looks like this:

```
[loggerize] [INFO] [2017-06-27 14:46:51 +0200] HTTP/1.1 POST 200 127.0.0.1 /home fooo=bar&password=foo 18 {"foo"=>"bar", "location"=>{"city"=>"Paris"}} 0.004076932
```

:warning: **We need to decide on a format for them and the query parameters**

I personally think that leaving query and body parameters separated is useful, please note that we can access the query parameters as a hash too.

## Filtering parameters

:question: Could a utility like this be useful somewhere else / in another context?
:question: Should it be only for the body parameters (given as a hash) or any string given to the logger? (second solution would be way more complex and could be error prone)
:question: How much complex filtering do we want to support (ie: ability to filter `credit_card[number]` but not `random[number]`)
:question: What kind of test can a dev write to ensure that the parameter is indeed filtered?

jodosha:
@TiteiKo 
> I personally think that leaving query and body parameters separated is useful, please note that we can access the query parameters as a hash too.

I'd love to hear more about this.

Given both the Rack env key that we use (`"router.params"`) and action's `params` object don't distinguish between the params coming from the route interpolation (eg `/books/:id`) and from the POST/PATCH payload, why would we separate those during the logging phase?

Another aspect that is worth noting, is about params precedence. If `:id` is both available as route interpolation (again, `/books/:id`) and from the PATCH payload, **the one from the route takes the precedence**. If we separate the logging for these two sources (route interpolation and payload), it can be misleading for a developer that can see `:id` twice, but only one was available in action's `params`.

To be the bottom line is to try to log exactly what's available in `params`.

jodosha:
@TiteiKo 

> ❓ Could a utility like this be useful somewhere else / in another context?

As of `1.0.1`, `hanami-utils` uses `transproc` for `Utils::Hash` transformations. The migration is not completed yet, but we can use [`map_values`](http://www.rubydoc.info/gems/transproc/Transproc%2FHashTransformations.map_values) for this purpose.

> ❓ Should it be only for the body parameters (given as a hash) or any string given to the logger? (second solution would be way more complex and could be error prone)

While it's technically possible to transform the query string into a hash and then to apply the filtering (via [`Rack::QueryParser`](http://www.rubydoc.info/gems/rack/Rack/QueryParser)), I'd limit the filtering for now to the params, not to the query string.

> ❓ How much complex filtering do we want to support (ie: ability to filter `credit_card[number]` but not `random[number]`)

Good question. 😄  Also, which syntax do we want to make available for this feature? Eg. `Logger.new(filter: [:password])` or `Logger.new(filter: "credit_card[number]")`, or `Logger.new(filter: "credit_card.number")`?

I prefer to support the first and the last (if we can). The `credit_card[number]` syntax resembles the syntax that we have for HTML form structure, but we shouldn't forget about SQL queries in the future. So the idea is to decouple from HTML forms towards an agnostic syntax.

>❓ What kind of test can a dev write to ensure that the parameter is indeed filtered?

Something like this

Minitest:

```ruby
out, = capture_io do
  Hanami.logger.info({password: "secret"})
end

out.must_equal('{password: "[FILTERED]"}')
```

RSpec:

```ruby
expect { Hanami.logger.info({password: "secret"}) }.to output('{password: "[FILTERED]"}').to_stdout
```

TiteiKo:
See #225

Issue #208, opened 2017-04-11
classify sometimes falsly classifies strings


choallin:
If I have an already classified string with a namespace in it, classify destroys that string. e.g.
```
"AwesomeProject::Namespace".classify # ==> "Awesome::ProjectNamespace"
```

jodosha:
Fixed by #209 and 9f150a9d963c753d88f579cd1f316915cc551ba6

PR #209, opened 2017-04-11
Fix wrong classify behavior if the string is already classified

choallin:
See issue #208

### DIFF
--- a/lib/hanami/utils/string.rb
+++ b/lib/hanami/utils/string.rb
@@ -141,6 +141,7 @@ module Hanami
       #   string = Hanami::Utils::String.new 'hanami_utils'
       #   string.classify # => 'HanamiUtils'
       def classify
+        return self.class.new to_s if classified?
         words = underscore.split(CLASSIFY_WORD_SEPARATOR).map!(&:capitalize)
         delimiters = scan(CLASSIFY_WORD_SEPARATOR)
 
@@ -149,6 +150,10 @@ module Hanami
         end
 
         self.class.new words.zip(delimiters).join
+      end
+
+      def classified?
+        split(NAMESPACE_SEPARATOR).reduce(false) {|memo, str| str.match(/\A[A-Z][a-zA-Z_]*/).to_s.length == str&.length }
       end
 
       # Return a downcased and underscore separated version of the string
--- a/lib/hanami/utils/string.rb
+++ b/lib/hanami/utils/string.rb
@@ -66,6 +66,12 @@ module Hanami
       # @api private
       DASHERIZE_SEPARATOR = '-'.freeze
 
+      # Regexp for #classified
+      #
+      # @since 1.0.x
+      # @api private
+      CLASSIFIED_CHECK = /\A[A-Z]\w*/
+
       # Regexp for #classify
       #
       # @since 0.3.4
@@ -141,7 +147,7 @@ module Hanami
       #   string = Hanami::Utils::String.new 'hanami_utils'
       #   string.classify # => 'HanamiUtils'
       def classify
-        return self.class.new to_s if classified?
+        return self if classified?
         words = underscore.split(CLASSIFY_WORD_SEPARATOR).map!(&:capitalize)
         delimiters = scan(CLASSIFY_WORD_SEPARATOR)
 
@@ -152,8 +158,22 @@ module Hanami
         self.class.new words.zip(delimiters).join
       end
 
+      # Return true if the string is a valid class name
+      #
+      # @return [Boolean]
+      #
+      # @since 1.0.x
+      #
+      # @example
+      #   require 'hanami/utils/string'
+      #
+      #   string = Hanami::Utils::String.new 'hanami_utils'
+      #   string.classified? # => false
+      #
+      #   string = Hanami::Utils::String.new 'HanamiUtils'
+      #   string.classified? # => true
       def classified?
-        split(NAMESPACE_SEPARATOR).reduce(false) {|memo, str| str.match(/\A[A-Z][a-zA-Z_]*/).to_s.length == str&.length }
+        split(NAMESPACE_SEPARATOR).all? {|str| str.match(CLASSIFIED_CHECK).to_s.length == str.to_s.length }
       end
 
       # Return a downcased and underscore separated version of the string
--- a/test/string_test.rb
+++ b/test/string_test.rb
@@ -1,4 +1,4 @@
-require 'test_helper'
+require_relative 'test_helper'
 require 'hanami/utils'
 require 'hanami/utils/string'
 
@@ -70,6 +70,8 @@ describe Hanami::Utils::String do
       Hanami::Utils::String.new('hanami/router').classify.must_equal('Hanami::Router')
       Hanami::Utils::String.new('hanami::router').classify.must_equal('Hanami::Router')
       Hanami::Utils::String.new('hanami::router/base_object').classify.must_equal('Hanami::Router::BaseObject')
+      Hanami::Utils::String.new('AwesomeProject::Namespace').classify.must_equal('AwesomeProject::Namespace')
+      Hanami::Utils::String.new('AwesomeProject::SubNamespace').classify.must_equal('AwesomeProject::SubNamespace')
     end
 
     it 'returns a classified string from symbol' do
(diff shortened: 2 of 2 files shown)


lib/hanami/utils/string.rb
```diff
@@ -151,6 +152,10 @@ def classify
         self.class.new words.zip(delimiters).join
       end
 
+      def classified?
+        split(NAMESPACE_SEPARATOR).reduce(false) {|memo, str| str.match(/\A[A-Z][a-zA-Z_]*/).to_s.length == str&.length }
```
davydovanton: I think it's so complicated method and we need to simplify it.

That's why I suggest:
1. Set regexp to constant
2. Instead `#reduce` you can use [`#all?`](https://apidock.com/ruby/Enumerable/all%3F) method
3. I think you can use `/\w+/` instead `/[a-zA-Z_]/`

lib/hanami/utils/string.rb
```diff
@@ -141,6 +141,7 @@ def capitalize
       #   string = Hanami::Utils::String.new 'hanami_utils'
       #   string.classify # => 'HanamiUtils'
       def classify
+        return self.class.new to_s if classified?
```
davydovanton: I think we can return `self` here because we don't mutate state 🤔

test/string_test.rb
```diff
@@ -85,6 +87,20 @@
...
+    it 'correctly recognizes a classified String' do
+      Hanami::Utils::String.new('hanami').classified?.must_equal(false)
+      Hanami::Utils::String.new('Hanami_router').classified?.must_equal(true)
+      Hanami::Utils::String.new('hanami-router').classified?.must_equal(false)
+      Hanami::Utils::String.new('Hanami/router').classified?.must_equal(false)
+      Hanami::Utils::String.new('Hanami/router').classified?.must_equal(false)
+      Hanami::Utils::String.new('Hanami/Router').classified?.must_equal(false)
+      Hanami::Utils::String.new('HanamiRouter').classified?.must_equal(true)
+      Hanami::Utils::String.new('AwesomeProject::Namespace').classified?.must_equal(true)
+      Hanami::Utils::String.new('AwesomeProject::SubNamespace').classified?.must_equal(true)
```
davydovanton: Really strange logic for `Hanami_router` and `Hanami/Router`, but it just my opinion 😃

lib/hanami/utils/string.rb
choallin: You are right, self should be fine here!

test/string_test.rb
choallin: Since ```Hanami_router``` is a valid class name it should be true

choallin:
I will also comment the ```classified?``` method

lib/hanami/utils/string.rb
mereghost: Also, classes can legally have numbers on it's names, so `\w` would be more suitable as @davydovanton said. =)

mereghost:
Oh, @choallin could you fix the rubocop warning so the CI build will pass? =)

artofhuman:
Should we write tests with `RSpec` instead `Minitest`. In master we alread run all tests by `RSpec`
https://github.com/hanami/utils/blob/master/Rakefile#L24

davydovanton:
awesome catch @jodosha @artofhuman 👏

test/string_test.rb
```diff
@@ -85,6 +87,20 @@
     end
   end
 
+  describe '#classified' do
```
jodosha: @choallin You introduced unit tests for `#classified?`, but not for the original issue, which was related to `#classify`.

lib/hanami/utils/string.rb
```diff
@@ -141,6 +147,7 @@ def capitalize
       #   string = Hanami::Utils::String.new 'hanami_utils'
       #   string.classify # => 'HanamiUtils'
       def classify
+        return self if classified?
```
jodosha: @choallin This looks like a workaround that doesn't fix the real problem in the following lines. Would you like to check why it doesn't work properly without adding this `#classfied?` guard?

---

More on the guard. Guards are useful if they are inexpensive. This is not the case: `#classified?` requires a lot of computations.

lib/hanami/utils/string.rb
```diff
@@ -151,6 +158,24 @@ def classify
...
+      #
+      # @example
+      #   require 'hanami/utils/string'
+      #
+      #   string = Hanami::Utils::String.new 'hanami_utils'
+      #   string.classified? # => false
+      #
+      #   string = Hanami::Utils::String.new 'HanamiUtils'
+      #   string.classified? # => true
+      def classified?
```
jodosha: If we decide to go with `#classified?` this should be a private method and `@api private`.

test/string_test.rb
choallin: @jodosha I added the lines 73 & 74 to test the original issue. Should I make a new test case for that?

lib/hanami/utils/string.rb
choallin: @jodosha I will take a look at the regex/method again if I can make it without it.

test/string_test.rb
jodosha: @choallin Nope, sorry. I missed that lines. That tests are enough.

TiteiKo:
I think one of the current tests is wrong, making everything way more complex...

```ruby
expect(Hanami::Utils::String.new('hanami-router').classify).to eq('Hanami::Router')
```

As 'hanami-router' underscored equals 'hanami_router' and not 'hanami/router', and as `#classify` relies on `#underscore`, then  it should produce 'HanamiRouter' instead of 'Hanami::Router'.

If we consider that this test is incorrect, then fixing the `#classify` method becomes way simpler:

```diff
diff --git a/lib/hanami/utils/string.rb b/lib/hanami/utils/string.rb
index 5ee64fa..4da25ab 100644
--- a/lib/hanami/utils/string.rb
+++ b/lib/hanami/utils/string.rb
@@ -142,7 +142,7 @@ module Hanami
       #   string.classify # => 'HanamiUtils'
       def classify
         words = underscore.split(CLASSIFY_WORD_SEPARATOR).map!(&:capitalize)
-        delimiters = scan(CLASSIFY_WORD_SEPARATOR)
+        delimiters = underscore.scan(CLASSIFY_WORD_SEPARATOR)

         delimiters.map! do |delimiter|
           delimiter == CLASSIFY_SEPARATOR ? EMPTY_STRING : NAMESPACE_SEPARATOR
diff --git a/spec/unit/hanami/utils/string_spec.rb b/spec/unit/hanami/utils/string_spec.rb
index 8138b85..3d86001 100644
--- a/spec/unit/hanami/utils/string_spec.rb
+++ b/spec/unit/hanami/utils/string_spec.rb
@@ -64,16 +64,18 @@ RSpec.describe Hanami::Utils::String do
     it 'returns a classified string' do
       expect(Hanami::Utils::String.new('hanami').classify).to eq('Hanami')
       expect(Hanami::Utils::String.new('hanami_router').classify).to eq('HanamiRouter')
-      expect(Hanami::Utils::String.new('hanami-router').classify).to eq('Hanami::Router')
+      expect(Hanami::Utils::String.new('hanami-router').classify).to eq('HanamiRouter')
       expect(Hanami::Utils::String.new('hanami/router').classify).to eq('Hanami::Router')
       expect(Hanami::Utils::String.new('hanami::router').classify).to eq('Hanami::Router')
       expect(Hanami::Utils::String.new('hanami::router/base_object').classify).to eq('Hanami::Router::BaseObject')
+      expect(Hanami::Utils::String.new('AwesomeProject::Namespace').classify).to eq('AwesomeProject::Namespace')
+      expect(Hanami::Utils::String.new('AwesomeProject::SubNamespace').classify).to eq('AwesomeProject::SubNamespace')
     end

     it 'returns a classified string from symbol' do
       expect(Hanami::Utils::String.new(:hanami).classify).to eq('Hanami')
       expect(Hanami::Utils::String.new(:hanami_router).classify).to eq('HanamiRouter')
-      expect(Hanami::Utils::String.new(:'hanami-router').classify).to eq('Hanami::Router')
+      expect(Hanami::Utils::String.new(:'hanami-router').classify).to eq('HanamiRouter')
       expect(Hanami::Utils::String.new(:'hanami/router').classify).to eq('Hanami::Router')
       expect(Hanami::Utils::String.new(:'hanami::router').classify).to eq('Hanami::Router')
     end
```

PR #210, opened 2017-04-25
Introduce `Utils::Hash.symbolize` and `.deep_symbolize`

jodosha:
Partial implementation of this proposal: https://discourse.hanamirb.org/t/transform-utils-string-and-utils-hash-to-modules/336

Ref hanami/model#394
Ref hanami/model#395

### DIFF
--- a/hanami-utils.gemspec
+++ b/hanami-utils.gemspec
@@ -18,6 +18,8 @@ Gem::Specification.new do |spec|
   spec.require_paths = ['lib']
   spec.required_ruby_version = '>= 2.3.0'
 
+  spec.add_dependency 'transproc', '~> 1.0'
+
   spec.add_development_dependency 'bundler', '~> 1.6'
   spec.add_development_dependency 'rake',    '~> 11'
   spec.add_development_dependency 'rspec',   '~> 3.5'
--- a/lib/hanami/utils/hash.rb
+++ b/lib/hanami/utils/hash.rb
@@ -1,4 +1,5 @@
 require 'hanami/utils/duplicable'
+require 'transproc'
 
 module Hanami
   module Utils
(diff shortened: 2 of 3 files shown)


hanami-utils.gemspec
```diff
@@ -18,6 +18,8 @@ Gem::Specification.new do |spec|
   spec.require_paths = ['lib']
   spec.required_ruby_version = '>= 2.3.0'
 
+  spec.add_dependency 'transproc', '~> 1.0'
```
davydovanton: WDYT can we use transproc as an optional dependency?

mereghost: Seems to be the easiest way to not break compatibility right now.

AlfonsoUceda:
LGTM 👍

hanami-utils.gemspec
jodosha: @davydovanton I'm torn about this. Generally speaking, I'd like the idea of having no deps for `utils`.

On the other hand, because `Utils::Hash` (and `Utils::String`) are widely used by our other gems, then developers will need to take care to manually install `transproc` and to maintain the dependency level that we use.

Example: I'm using only `hanami-controller` and `hanami-router` in a project of mine. So my `Gemfile` looks like this:

```ruby
# ...
gem 'hanami-controller', '~> 1.0'
gem 'hanami-router', '~> 1.0'
```

There is not even `hanami-utils` in my `Gemfile`. It would be odd to add `transproc`, as soft-dependency for a dependency that I don't even list in my `Gemfile`.

An additional maintenance burden for us would be the version of `transproc` that devs will use. Let's get back to our example, I add the gem:

```ruby
# ...
gem 'transproc'
```

We don't have any guarantee that a developer will pick the right version. That `Gemfile` instruction means "install the latest stable version of `transproc`", which may not match what we need in `utils`. Right now we enforce `transproc` 1.x series, but we can't prevent a developer to do `bundle update` and get 2.0.

Please note that `transproc` 2.0 isn't out, it was just an example.

---

Punchline is: it's probably better to add that direct dep in gemspec.

jodosha:
@TiteiKo WDYT?

hanami-utils.gemspec
davydovanton: I got it, thanks for the detailed response 👏

PR #211, opened 2017-04-29
Introduce new interactor format

TiteiKo:
ATM, the parameters for the execution are send at the initialization of the instance. This behavior poses 2 main issues:
* it is really hard to make Dependency Injections
* there is no way to reuse an Interactor instance with other parameters

This PR aims at introducing a new format that would solve the previously mentioned issues, all the while still supporting the previous format.

New format example:

```ruby
class Hello
  include Hanami::Interactor
  expose :hello

  def call(target:)
    @hello = "Hello #{target}"
  end

  private

  def valid?(target:)
    !target == 'me'
  end
end

hello = Hello.new
hello.call(target: 'world') # => valid?: true, hello: 'Hello world'
hello.call(target: 'me') # => valid?: false
```

TODO:

- [x] prove that it can work
- [x] write all the current tests for both formats
- [x] write all the documentation

### DIFF
--- a/lib/hanami/interactor.rb
+++ b/lib/hanami/interactor.rb
@@ -353,18 +353,18 @@ module Hanami
       #   end
       #
       #   Signup.new.call # => NoMethodError
-      def call(**attrs)
+      def call(*args, **kwargs)
         @__result = ::Hanami::Interactor::Result.new
-        _call(attrs) { super }
+        _call(*args, **kwargs) { super }
       end
 
       private
 
       # @api private
       # @since x.x.x
-      def _call(**attrs)
+      def _call(*args, **kwargs)
         catch :fail do
-          validate!(attrs)
+          validate!(*args, **kwargs)
           yield
         end
 
@@ -372,8 +372,8 @@ module Hanami
       end
 
       # @since x.x.x
-      def validate!(**attrs)
-        fail! unless valid?(attrs)
+      def validate!(*args, **kwargs)
+        fail! unless valid?(*args, **kwargs)
       end
     end
 
--- a/lib/hanami/interactor.rb
+++ b/lib/hanami/interactor.rb
@@ -145,15 +145,14 @@ module Hanami
       super
 
       base.class_eval do
-        prepend Interface
-        extend  ClassMethods
+        extend ClassMethods
       end
     end
 
-    # Interactor interface
+    # Interactor legacy interface
     #
     # @since 0.3.5
-    module Interface
+    module LegacyInterface
       # Initialize an interactor
       #
       # It accepts arbitrary number of arguments.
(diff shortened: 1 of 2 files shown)


cllns:
🎉 very excited about this :)

gizotti:
This is awesome. I bumped into the dependency injection problem last week.

jodosha:
@TiteiKo Thanks for taking care of this. 💯  I changed the base branch to `develop` because this should be part of `utils` 1.1.0.

AlfonsoUceda:
@TiteiKo awesome and fantastic work, keep working on this ;)

lib/hanami/interactor.rb
```diff
@@ -476,6 +500,16 @@ def self.extended(interactor)
       end
     end
 
+    def method_added(method_name)
+      return unless method_name == :call
```
mereghost: isn't it work to bubble the hook up by calling super?

TiteiKo: Yep indeed, thought of it afterwards, will do

mereghost:
Quick question: does marking the `LegacyInterface` as deprecated be beneficial to us in the long run?

TiteiKo:
@mereghost Yes it would, but I'd like to wait until we have some documentation of the new syntax so we can help people migrate by putting a link in the deprecation.

mereghost:
@hanami/core can we :shipit: ?

lib/hanami/interactor.rb
```diff
@@ -264,6 +263,118 @@ def initialize(*args)
...
+      #   require 'hanami/interactor'
+      #
+      #   class Signup
+      #     include Hanami::Interactor
+      #
+      #     # Method #call is not defined
+      #   end
+      #
+      #   Signup.new.call # => NoMethodError
+      def call(**attrs)
```
TiteiKo: Could be `call(*args, **attrs)` to really allow any type of params combination... 
```ruby
# will fail with (**attrs) but not with (*args, **attrs)
MyInteractor.new.call('foo', bar: 'baz')
MyInteractor.new.call('foo')
```

WDYT should I make the change?

cllns: What are the disadvantages? Seems like we should be as flexible as possible, IMO

mereghost: I think it's more of a matter of what we want to enforce. If we like the more "keyword-ish" way (which I really like, coming from Smalltalk) or just allow anything.

TiteiKo:
@cllns is right, there is no downside and we should be as open as possible. I also agree with @mereghost and prefer kwargs, but maybe we should push this as a best practice and not enforce it...

Anyway, interactors now accept args and kwargs in call :)

Good to merge if good for you

TiteiKo:
@jodosha thanks! :green_heart:

Issue #213, opened 2017-06-13
Documentation issue with persist method


jaymiejones86:
I encountered an issue when following the documentation for using the [Interactor Interface](http://www.rubydoc.info/gems/hanami-utils/Hanami/Interactor/Interface) where the repository is being called with the `persist` method which in my Hanami application would result in a `NoMethodError` exception.

To ensure it was not own code causing the issue, I have created a repo to mimic the documentation and written a passing test to confirm the behaviour of the NoMethodError, which can be found at https://github.com/jaymiejones86/hanami-utils-interactor-persist-bug

I thought of bringing this issue up here prior to issuing a pull request incase there is something I have missed as I am still very new to the framework and the various utils.

*Environment details*
- Ruby 2.3.1
- Hanami 1.0
- Mac OSX 10.12.5

cllns:
The `persist` class method hasn't existed for a while. First, we moved from class methods to object methods. Then, we changed the back-end of `hanami-model` and removed `persist` entirely. These docs just never got updated. See the methods on Repositories here: https://github.com/hanami/model#repositories

Feel free to open a PR changing the lines with `persist` to `UserRepository.new.create(@params)`, and changing the line in the initializer to `@user = nil`.

cllns:
@jaymiejones86 Would you like to open the PR, or should I do it? 🌸

jaymiejones86:
@cllns I will create the PR today. Ran out of time since your last comment.

cllns:
Closed via #215! 🎉

🌸

PR #215, opened 2017-06-15
Update Documentation To Not Use persist Method

jaymiejones86:
Update documentation for the persisting of the repository to use the
create method instead of the persist method due to the persist method
not existing for quite some time.

The example could confuse a user trying to implement the pattern causing
false issues being posted or time loss.

Relates to issue #213


jaymiejones86:
@davydovanton thanks for reviewing and approving 👍

PR #216, opened 2017-06-19
Test frozen input for Utils::Hash#initialize

badeball:
This test does currently not pass and is merely a suggestion for defining its behavior. Personally, I think it's a tad bit icky to modify arguments and I would like to be able to duplicate frozen objects. What do you think?

### DIFF
--- a/spec/unit/hanami/utils/hash_spec.rb
+++ b/spec/unit/hanami/utils/hash_spec.rb
@@ -176,6 +176,13 @@ RSpec.describe Hanami::Utils::Hash do
       expect(hash.to_h).to eq(input_to_hash.to_hash)
     end
 
+    it 'accepts frozen values' do
+      pending
+
+      expect { Hanami::Utils::Hash.new({}.freeze) }
+        .to_not raise_error
+    end
+
     it "raises error when object doesn't implement #to_hash" do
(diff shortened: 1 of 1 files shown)


davydovanton:
hey, thanks for PR but all tests passed 🤔

badeball:
Yeah, this is the error generated.

```
irb(main):016:0> Hanami::Utils::Hash.new({}.freeze)
RuntimeError: can't modify frozen Hash
	from /home/jonas/.gem/ruby/2.3.3/gems/hanami-utils-1.0.0/lib/hanami/utils/hash.rb:48:in `default_proc='
	from /home/jonas/.gem/ruby/2.3.3/gems/hanami-utils-1.0.0/lib/hanami/utils/hash.rb:48:in `initialize'
	from (irb):16:in `new'
	from (irb):16
```

mereghost:
Not sure if we should mess with frozen values. =/ My gut reaction would be, leave those alone. :P

TiteiKo:
From what I understand, this error means that we mutate the hash given as parameter, which we shouldn't... And from what I see, we're doing it without reason to do it. 

@badeball Could you look at #219 see if it answers your issue? :)

jodosha:
@badeball Thanks for this PR. I'm closing this in favor of #219 👍

Issue #217, opened 2017-06-21
Inflector mistake on words like `phase`


sovetnik:
```ruby
Hanami::Utils::Inflector.singularize('phases')
=> "phas"
```

Can be hardcodeв by adding exception in initializer (thanks @TiteiKo)
```ruby
Hanami::Utils::Inflector.inflections do
  exception 'phase', 'phases'
end
```

jodosha:
@sovetnik We've just released `hanami-utils-1.0.2` with this fix. Thanks 👍

PR #218, opened 2017-06-22
Add 'phase' word to rules for singulars

davydovanton:
Fixes https://github.com/hanami/utils/issues/217

/cc @hanami/core

### DIFF
--- a/lib/hanami/utils/inflector.rb
+++ b/lib/hanami/utils/inflector.rb
@@ -269,7 +269,8 @@ module Hanami
         'species'     => 'species',
         'police'      => 'police',
         # fallback
-        'hives'       => 'hive'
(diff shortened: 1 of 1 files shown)


TiteiKo:
In the inflector, there's a rule that says that plural words ending in `ses` should be singularized using only `s`...
`phase` doesn't look like an exception, maybe our `ses -> s` rule isn't precise enough? 

From what I can see looking at singulars of english words ending in `es`:
* `xes -> x`
* `ses -> se`
* `res -> r`
* `uses -> us`
* `sses -> ss`
* `ches -> ch`

jodosha:
I'm merging this PR and implementing the missing specs into a separated commit.

PR #220, opened 2017-07-05
Introduce `Utils::Hash.stringify`

marionschleifer:


TiteiKo:
Why?

marionschleifer:
@TiteiKo I watched the build and something was marked in red (even though the whole build passed), so I changed it. Should I change it back?

TiteiKo:
Why was on the "why change it to a class method"

This change is not backward compatible, meaning that people using `Hanami::Utils::Hash.new(foo: 'bar').stringify!` will now get an error :(

Because we're good people, we don't make backward incompatible changes in minor versions, only between major versions (so this would be for 2.0)

What you could do is:
* keep the current test and code
* add your new test
* add your new code
* see if one can use the other

On a more global note, @hanami/core do we want to make this method a class method? (I personally don't think so)

davydovanton:
First of all, great work! 🌟

So, I think we can improve these changes. And I want to suggest two things:
1. Lets use [deprication](https://github.com/hanami/utils/blob/master/lib/hanami/utils/deprecation.rb) module instead of replacing method. (We have to prepare our users for new changes).
2. We can create `ClassMethods` module and put all class methods here. Why i think it's good: we will have all new logic in one place and it will be easy to add new class methods/remove instance methods in future

PR #226, opened 2017-08-09
Introduce `Hanami::Utils::String.transform`

jodosha:
## Proposal

Apply the given transformation(s) to `input`

It performs a pipeline of transformations, by applying the given functions from `Hanami::Utils::String` and `::String`

The transformations are applied with the given order.

It doesn't mutate the input, unless a developer uses destructive methods from `::String`

## Usage

```ruby
require "hanami/utils/string"

# Apply transformations from `Hanami::Utils::String`
Hanami::Utils::String.transform("hanami/utils", :underscore, :classify)
  # => "Hanami::Utils"

# Apply transformations from `::String`
Hanami::Utils::String.transform("Hanami", :downcase, :reverse)
  # => "imanaH"

# Apply transformations with arguments
Hanami::Utils::String.transform("Hanami::Utils::String", [:gsub, /[aeiouy]/, "*"], :demodulize)
  # => "H*n*m*"

# Apply proc transformations
Hanami::Utils::String.transform("Hanami", ->(s) { s.upcase })
  # => "HANAMI"
```

## Why?

In our codebase we have chained transformations like:

```ruby
Hanami::Utils::String.new(string).underscore.classify
```

or

```ruby
Hanami::Utils::String.new(self.class.name).demodulize.downcase
```

We're moving towards the removal of `Hanami::Utils::String` as a type, but we'll keep it as a module to perform string transformations. (See #224 )

That code snippets above should be rewritten as:

```ruby
result = Hanami::Utils::String.underscore(string)
result = Hanami::Utils::String.classify(result)
result
```

or

```ruby
result = Hanami::Utils::String.demodulize(self.class.name)
result = result.downcase
result
```

While these multiple liners do their work, they are not elegant like:

```ruby
Hanami::Utils::String.transform(string, :underscore, :classify)
```

or

```ruby
Hanami::Utils::String.transform(self.class.name, :demodulize, :downcase)
```

## Implementation details

Building a transformation chain is CPU intensive. For this reason they are stored in a thread-safe cache.

This cache has a boost that goes from `1.32x` (basic case) to `3.23x` (more complex cases).

Please have a look at https://gist.github.com/jodosha/abadd8e42aeaa1b0944f5bac3518bac9

### DIFF
--- a/lib/hanami/utils/string.rb
+++ b/lib/hanami/utils/string.rb
@@ -1,4 +1,6 @@
 require 'hanami/utils/inflector'
+require 'transproc'
+require 'concurrent/map'
 
 module Hanami
   module Utils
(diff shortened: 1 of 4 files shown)


lib/hanami/utils/string.rb
```diff
@@ -72,6 +74,97 @@ class String # rubocop:disable Metrics/ClassLength
...
+      #   Hanami::Utils::String.transform("Cherry", -> { "blossom" }))
+      #     # => ArgumentError: wrong number of arguments (given 1, expected 0)
+      #
+      # rubocop:disable Metrics/MethodLength
+      # rubocop:disable Metrics/AbcSize
+      def self.transform(input, *transformations)
+        fn = @__transformations__.fetch_or_store(transformations.hash) do
+          compose do |fns|
+            transformations.each do |transformation|
+              tr, *args = transformation
```
davydovanton: WDYT maybe is better to wrap this block to the private method?

jodosha: I think we're good 😉 👍

davydovanton: or we can try to move this logic to separate class 🤔

jodosha:
Travis CI is really slow on today. I'm taking the responsibility of merging this while CI is still running. I verified that the build passes for JRuby as well.

In case the CI will be still broken because of this merge, I'll take care of fixing it right after.

---

I'm forcing this because, I have some free time and wanna prepare the gems for `1.1.0.beta1`.

PR #228, opened 2017-08-28
Fix the singular of exercises (exercise rather than exercis).

malin-as:
I encountered an error with the singularisation rule of `exercises`, so I fixed it by adding an exception and a test for it.

### DIFF
--- a/lib/hanami/utils/inflector.rb
+++ b/lib/hanami/utils/inflector.rb
@@ -268,7 +268,8 @@ module Hanami
         'police'      => 'police',
         # fallback
         'hives'       => 'hive',
-        'phases'      => 'phase'
+        'phases'      => 'phase',
+        'exercises'   => 'exercise'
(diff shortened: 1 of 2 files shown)


cllns:
Thanks for the contribution! ✨🌸

For reference, without this patch:

Hanami::Utils::Inflector.singularize("exercises")
=> "exercis"
Can you fix a couple things:

- Add a trailing comma to the line you added. This isn't a big deal, but it'll keep future diffs cleaner (like in this case, the previous entry didn't have a comma)

- Can you add a link to this issue in a comment, like "album" and "phase" entries?

Thanks so much!

malin-as:
@cllns Sorry I got busy and didn't have time to do the small fixes, should I open a bug and reference it in a new pull request?

cllns:
@malin-as don't worry about it :)

PR #229, opened 2017-08-31
fix the pluralization of area

johnnoone:


cllns:
Thanks for the contribution! ✨🌸 

For reference, without this patch:
```
> Hanami::Utils::Inflector.pluralize("area")
=> "area"
```

Can you fix a couple things:
- Add a trailing comma to the line you added. This isn't a big deal, but it'll keep future diffs cleaner (like in this case, the previous entry didn't have a comma, so that 

- In the fixtures file, can you move the added `album` entry?

- Can you add a link to this issue in a comment, like album has?

Thanks so much!

jodosha:
@johnnoone Thank you very much, I'm merging this. 👍

PR #223, opened 2017-07-26
Introduce `Utils::Files`

jodosha:
This is a set of helpers that we were using since long time in `hanami` test suite.

With the new CLI for `hanami`, based on `hanami-cli`, we needed to extract that helpers from the test suite and promote it as production code.

Ref https://github.com/hanami/hanami/pull/808

PR #231, opened 2017-09-14
logger: creates folder when doesn't exist

AlfonsoUceda:
I've noticed when you create a hanami app and add a stream in your development logger, for example a path 'log/development.log', it will fails because log folder isn't created. IMO logger should be in charge of create the path.

What do you think?

### DIFF
--- a/lib/hanami/logger.rb
+++ b/lib/hanami/logger.rb
@@ -5,6 +5,7 @@ require 'hanami/utils/string'
 require 'hanami/utils/json'
 require 'hanami/utils/hash'
 require 'hanami/utils/class_attribute'
+require 'hanami/utils/files'
 
 module Hanami
   # Hanami logger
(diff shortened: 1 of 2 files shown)


cllns:
You can use `Hanami::Utils::Files` for this, from #223 :) https://github.com/hanami/utils/blob/develop/lib/hanami/utils/files.rb#L21-L32

AlfonsoUceda:
@cllns as far as I know, ruby logger class creates the file, we only need to create the path, is that right?

AlfonsoUceda:
Before merging I have to create a PR in hanami.

AlfonsoUceda:
Tested in a fresh applications using `stream: "log/development.log"`and it works 👌

lib/hanami/logger.rb
```diff
@@ -453,7 +454,14 @@ def _format(hash)
     #   logger.info "Hello World"
     #
     #   # => {"app":"Hanami","severity":"DEBUG","time":"2017-03-30T13:57:59Z","message":"Hello World"}
-    def initialize(application_name = nil, *args, stream: $stdout, level: DEBUG, formatter: nil, filter: []) # rubocop:disable Metrics/ParameterLists
+    # rubocop:disable Lint/HandleExceptions
+    # rubocop:disable Metrics/ParameterLists
+    def initialize(application_name = nil, *args, stream: $stdout, level: DEBUG, formatter: nil, filter: [])
+      begin
+        Utils::Files.mkdir_p(stream)
+      rescue TypeError
```
jodosha: @AlfonsoUceda Are we trying to create the directory regardless of the `stream` type?

AlfonsoUceda: Yep, but I think that the reason of rescue

PR #232, opened 2017-09-22
Refactored Hash’s deep_symbolize! method to be able to symbolise nest…

rybex:
…ed array of hashes to keep consistency

### DIFF
--- a/lib/hanami/utils/hash.rb
+++ b/lib/hanami/utils/hash.rb
@@ -184,7 +184,11 @@ module Hanami
       def deep_symbolize!
         keys.each do |k|
           v = delete(k)
-          v = self.class.new(v).deep_symbolize! if v.respond_to?(:to_hash)
+          if v.respond_to?(:to_hash)
+            v = self.class.new(v).deep_symbolize!
+          elsif v.is_a?(Array)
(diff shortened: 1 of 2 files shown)


mereghost:
@rybex could you fix the rubocop warnings so it does not break the CI checks. =)

rybex:
@mereghost My bad. Fixed ;)

cllns:
Should we check for `is_a?(Array)`, or could we do `.respond_to?(:each)`, to see if it's Enumerable?

This would have a benefit of working on other Enumerable objects, like a `Set` of strings would symbolize the contents of the set.

Just a thought, there might be a good reason we don't want to do this. Hash-like objects are enumerable, but they're checked first, so we don't have to worry about them.

jodosha:
@rybex Thanks for this PR, but this can't be merged.

You changed the behavior of `Hanami::Utils::Hash#deep_symbolize!` (instance method), but you tested the behavior of `Hanami::Utils::Hash.deep_symbolize` (class method).

If you revert your changes to `lib/hanami/utils/hash.rb` the spec will pass. That means, the class method is already able to provide the behavior you are looking for.

Also, we're migrating away from the instance methods of `Hanami::Utils::Hash`.

I'm closing this, and I encourage you to use the class method instead. Thank you.

PR #239, opened 2017-10-27
Fix Files.write behavior

cllns:
The comments for `Hanami::Utils::Files.write` says:

```
# If the path already exists, it appends the contents.
```

The method did not do this. Instead, [it replaced the beginning of the file](https://github.com/hanami/cli/issues/22). 

Fixes part of https://github.com/hanami/hanami/issues/855.

### DIFF
--- a/lib/hanami/utils/files.rb
+++ b/lib/hanami/utils/files.rb
@@ -169,7 +169,7 @@ module Hanami
         content = ::File.readlines(path)
         content.unshift("#{line}\n")
 
-        rewrite(path, content)
+        write(path, content, overwrite: true)
       end
 
       # Adds a new line at the bottom of the file
@@ -188,7 +188,7 @@ module Hanami
         content = ::File.readlines(path)
         content << "#{contents}\n"
 
-        rewrite(path, content)
+        write(path, content, overwrite: true)
       end
 
       # Replace first line in `path` that contains `target` with `replacement`.
@@ -207,7 +207,7 @@ module Hanami
         content = ::File.readlines(path)
         content[index(content, path, target)] = "#{replacement}\n"
 
-        rewrite(path, content)
+        write(path, content, overwrite: true)
       end
 
       # Replace last line in `path` that contains `target` with `replacement`.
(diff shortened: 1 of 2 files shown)


AlfonsoUceda:
@cllns it seems good but try to create a branch in hanami with this utils branch and check if test build passes.

cllns:
> Is this a breaking change?
It's a breaking change but also a bug fix. The current behavior is really bad (only writing the beginning bits and ignoring the rest) so I see the bug fix portion as more important.
Three cases:
1) Writing empty files: no change
2) Overwriting a short file with something longer: breaking change, now appends
3) Overwriting a long file with something shorter: buggy behavior that no one would use

I think most people would use this for case 1. Think cases 2 and 3 are pretty unlikely (since you'd likely do both and it'd be broken).

> I'm not sure if we should follow that comment or not. 
> The one that says that content should be appended. If we go down that path, there is an overlap with .append.
> It doesn't solve hanami/hanami#855, as appending new code is not what we want.
You mentioned that Hanami CLI generators should use .rewrite instead. That would fix the bug, but because these file utils are gonna be used by third party devs to build CLI plugins, isn't it misleading to tell them to use .rewrite instead of .write? I mean, if a file doesn't exist, "rewriting" doesn't make sense..

Agreed. What if we change it so that there's just one method `write(overwrite: false)`? That way people could pass `overwrite: true`). Else, it'll throw a 'File Aready Exists' exception.
(We'd keep `rewrite` around until 2.0, of course, but we could deprecate it)

cllns:
OK pushed a couple commits to add a `FileAlreadyExistsError`, and an `overwrite` keyword argument, defaulted to false, to `write`. Also deprecated `rewrite` in favor of `write(..., overwrite: true)`.

I also implemented `FileAlreadyExistsError` and added `overwrite` keyword argument for `Files.cp`. If we go down that path, I think we should plan to make `Files.cp` **not** overwrite by default. (Breaking change, so couldn't happen before 2.0).

Let me know what y'all think about that 🌸

jodosha:
@cllns @AlfonsoUceda I'd like to have your feedback on the changes that I made to this PR.

 * `Utils::Files.write` is idempotent: it always outputs the same content, regardless of how many times is invoked.
  * `Utils::Files.rewrite` is deprecated, as there is an overlap with `.write`.
  * `Utils::Files.touch` is idempotent: it creates intermediary directories (if missing), and touch the file. It no longer erases its contents.
  * `Utils::Files.append` should be used to append contents to a file.
  * Never raise exceptions, as they don't make operations idempotent.

Thanks.

PR #242, opened 2017-10-28
Modify new interactor's argument list to handle some cases

tokichie:
Hi, I'm using hanami for my new project, and I love it. Thank you for a cool framework.
I've noticed that new interactor interface (from 1.1) cannot handle some argument patterns.
I propose a solution to handle such cases.

**case 1: no keyword arguments**
In this case, the actual arguments passed to #call is `['foo', 'bar', {}]`
This can be avoided to change #call args to `call(foo, bar, *)` but i think it is not cool 🤔 

```foo_interactor.rb
class FooInteractor
  include Hanami::Interactor
  def call(foo, bar)
  end
end
```

```
[3] pry(main)> interactor = FooInteractor.new
=> #<FooInteractor:0x007f9d2797ef98>
[4] pry(main)> interactor.call('foo', 'bar')
ArgumentError: wrong number of arguments (given 3, expected 2)
from (pry):4:in `call'
```

**case 2: no keyword arguments and an argument with `#to_hash` method**
In this case, the last argument is converted into Hash automatically if it implements `#to_hash`.
Because `Hanami::Entity` implements `#to_hash`, this case may happen often, and makes the behavior unexpected 😢 

```foo_interactor.rb
class FooInteractor
  include Hanami::Interactor
  def call(foo, bar)
    p foo, bar
  end
end

class Foo < Hanami::Entity; end
class Bar < Hanami::Entity; end
```

```
[8] pry(main)> foo = Foo.new(foo: 'foo')
=> #<Foo:0x007fb0513f2380 @attributes={:foo=>"foo"}>
[9] pry(main)> bar = Bar.new(bar: 'bar')
=> #<Bar:0x007fb051369378 @attributes={:bar=>"bar"}>
[10] pry(main)> interactor.call(foo, bar)
#<Foo:0x007fb0513f2380 @attributes={:foo=>"foo"}>
{:bar=>"bar"}
^^^^^^^^^^^^^ Bar is converted into a hash
=> #<Hanami::Interactor::Result:0x00007fb0512db1b8 @success=true @payload={}>
```

### DIFF
--- a/lib/hanami/interactor.rb
+++ b/lib/hanami/interactor.rb
@@ -351,18 +351,18 @@ module Hanami
       #   end
       #
       #   Signup.new.call # => NoMethodError
-      def call(*args, **kwargs)
+      def call(*args)
         @__result = ::Hanami::Interactor::Result.new
-        _call(*args, **kwargs) { super }
+        _call(*args) { super }
       end
 
       private
 
       # @api private
       # @since 1.1.0
-      def _call(*args, **kwargs)
+      def _call(*args)
         catch :fail do
-          validate!(*args, **kwargs)
+          validate!(*args)
           yield
         end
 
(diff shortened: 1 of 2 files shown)


cllns:
Thanks for this PR :)

@TiteiKo You worked on changing the args. What do you think about this change?

PR #246, opened 2017-12-13
0.9.x

augustoc100:
in the latest versions of ruby ​​that breaks

### DIFF
--- a/lib/hanami/utils/hash.rb
+++ b/lib/hanami/utils/hash.rb
@@ -44,9 +44,7 @@ module Hanami
       #
       #   hash.to_h # => { 'foo' => ['bar'] }
       def initialize(hash = {}, &blk)
-        require 'pry'
-        binding.pry
-        @hash = hash.to_hash
+        @hash = hash.to_h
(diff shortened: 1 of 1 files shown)


cllns:
Can you explain a little more about this PR please? Why does it target `0.9`?

augustoc100:
I just updated the hanami utils for version .9 and ruby ​​for 2.4.2 here, and he complains that the "to_hash" method does not exist, but if it changes to to_h it is in .7 version, it works

jodosha:
@augustoc100 Thanks for this PR. There is a difference between `#to_h` and `#to_hash`, that's why we have the latter.

Some objects respond to `#to_h`, but they aren't always able to return a `Hash`. The classic example is `Array`.

```ruby
[[:a, 1]].to_h
# => {:a=>1}

[:a, 1].to_h
# => TypeError: wrong element type Symbol (expected array)
```

Only the objects that implement `#to_hash` ensure to always return a `Hash`.

---

We can't merge this PR, because it would change how that method works.

PR #247, opened 2018-01-02
Official support for Ruby 2.5.0

jodosha:
This PR also removes the usage of `BigDecimal.new` in favor of `Kernel.BigDecimal`, as suggested by Ruby 2.5.0 warnings.


oana-sipos:
👍  good job, @jodosha!

PR #248, opened 2018-01-23
Fix String API docs

cllns:
While pairing programming with @oana-sipos today, I noticed some of the API docs for `Hanami::Utils::String` were unclear or wrong.

There was some usage of `String`, which wasn't clear in this context.

Now, all are either `::String` or `Hanami::Utils::String`

I also cleaned up a couple other API docs that I noticed were wrong (`nil` was given as a possible return value, but I don't think that's accurate, and `scan`'s comment didn't explain it).

### DIFF
--- a/lib/hanami/utils/string.rb
+++ b/lib/hanami/utils/string.rb
@@ -212,7 +212,7 @@ module Hanami
       #
       # @param input [::String] the input
       #
-      # @return [String] the transformed string
+      # @return [::String] the transformed string
       #
       # @since 1.1.0
       #
@@ -239,7 +239,7 @@ module Hanami
       #
       # @param input [::String] the input
       #
-      # @return [String] the transformed string
+      # @return [::String] the transformed string
       #
       # @since 1.1.0
       #
@@ -282,7 +282,7 @@ module Hanami
       #
       # @param input [::String] the input
       #
-      # @return [String] the transformed string
+      # @return [::String] the transformed string
       #
       # @since 1.1.0
       #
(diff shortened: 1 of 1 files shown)


oana-sipos:
Nice job, @cllns, glad you took action so quickly on this 🎉

lib/hanami/utils/string.rb
```diff
@@ -602,7 +602,7 @@ def singularize
 
       # Returns the hash of the internal string
       #
-      # @return [Fixnum]
+      # @return [Hash]
```
jodosha: I'm gonna fix this. It returns an `Integer` (previous versions of Ruby had the `Fixnum` class).

```ruby
"Hanami".hash
  # => 3882157824217614589
```

lib/hanami/utils/string.rb
```diff
@@ -657,9 +657,10 @@ def gsub(pattern, replacement = nil, &blk)
         end
       end
 
-      # Both forms iterate through str, matching the pattern
+      # Iterate through the string, matching the pattern.
+      # Either return all those patterns, or pass them to the block.
       #
-      # @return [String,nil]
+      # @return [::String]
```
jodosha: This is `Array<::String>`.

```ruby
irb(main):001:0> "Hanami".scan(/foo/)
=> []
irb(main):002:0> "Hanami".scan("foo")
=> []
irb(main):003:0> "Hanami".scan("Hanami")
=> ["Hanami"]
```

lib/hanami/utils/string.rb
```diff
@@ -602,7 +602,7 @@ def singularize
 
       # Returns the hash of the internal string
       #
-      # @return [Hash]
```
cllns: oops :)

PR #249, opened 2018-01-23
Colored output

oana-sipos:
This is a first implementation to render colored output in logs. This common work can be used both in `Hanami::CLI` and `Hanami::Utils::Logger`.

What do you think, @hanami/core? Any suggestions on the point we got so far? 

Lots of thanks for the guidance and pairing to @cllns, very nice to have done this!

### DIFF
--- a/lib/hanami/utils/string.rb
+++ b/lib/hanami/utils/string.rb
@@ -723,6 +723,48 @@ module Hanami
       def respond_to_missing?(m, include_private = false)
         @string.respond_to?(m, include_private)
       end
+
+      # Colorize output
+      # 8 colours available: black, red, green, yellow, blue, magenta, cyan, and gray
+      #
+      # @api private
+      # @since 1.2.beta1
+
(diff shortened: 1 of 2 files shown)


lib/hanami/utils/string.rb
```diff
@@ -723,6 +738,23 @@ def method_missing(m, *args, &blk)
       def respond_to_missing?(m, include_private = false)
         @string.respond_to?(m, include_private)
       end
+
+      # Colorize output
+      # 8 colors available: black, red, green, yellow, blue, magenta, cyan, and gray
+      #
+      # @api private
```
AlfonsoUceda: I think this method shouldn't be marked as private

AlfonsoUceda:
@oana-sipos good PR, can you fix it? After that I think we can merge it

jodosha:
Closing in favor of https://github.com/hanami/utils/pull/251

PR #251, opened 2018-02-13
Coloured logger

oana-sipos:
This is WIP. We still need to extract it in a module as @jodosha suggested.
CC @cllns please feedback it :) sorry it took me ages!

### DIFF
--- a/lib/hanami/logger.rb
+++ b/lib/hanami/logger.rb
@@ -475,18 +475,19 @@ module Hanami
     #   # => {"app":"Hanami","severity":"DEBUG","time":"2017-03-30T13:57:59Z","message":"Hello World"}
     # rubocop:disable Lint/HandleExceptions
     # rubocop:disable Metrics/ParameterLists
-    def initialize(application_name = nil, *args, stream: $stdout, level: DEBUG, formatter: nil, filter: [])
+    def initialize(application_name = nil, *args, stream: $stdout, level: DEBUG, formatter: nil, filter: [], colorize: false)
       begin
         Utils::Files.mkdir_p(stream)
       rescue TypeError
       end
 
       super(stream, *args)
+      colorize ||= tty?
 
       @level            = _level(level)
       @stream           = stream
       @application_name = application_name
-      @formatter        = Formatter.fabricate(formatter, self.application_name, filter, colorize: tty?)
+      @formatter        = Formatter.fabricate(formatter, self.application_name, filter, colorize: colorize)
     end
 
     # @since x.x.x
--- a/lib/hanami/logger.rb
+++ b/lib/hanami/logger.rb
@@ -203,15 +203,16 @@ module Hanami
       # @since 0.8.0
       # @api private
       def _format(hash)
-        [_line_front_matter(hash), _format_message(hash)].join(SEPARATOR)
+        [
+          _line_front_matter(hash[:app], hash[:severity], hash[:time]),
+          _format_message(hash)
+        ].join(SEPARATOR)
       end
 
       # @since x.x.x
       # @api private
-      def _line_front_matter(hash)
-        [hash[:app], hash[:severity], hash[:time]].map do |string|
-          "[#{string}]"
-        end.join(SEPARATOR)
+      def _line_front_matter(*args)
+        args.map { |string| "[#{string}]" }.join(SEPARATOR)
       end
 
       # @since x.x.x
--- a/lib/hanami/logger.rb
+++ b/lib/hanami/logger.rb
@@ -491,7 +491,7 @@ module Hanami
       end
 
       super(stream, *args)
-      colorize ||= tty?
+      colorize ||= colorize.nil? && tty? # By default, colorize if tty?
 
       @level            = _level(level)
       @stream           = stream
--- a/lib/hanami/logger.rb
+++ b/lib/hanami/logger.rb
@@ -216,7 +216,7 @@ module Hanami
         [
           _colored(hash[:app], color: :yellow),
           _colored(hash[:severity], color: :cyan),
-          _colored(hash[:time], color: :green),
+          _colored(hash[:time], color: :green)
         ].map { |string| "[#{string}]" }.join(SEPARATOR)
       end
 
--- a/lib/hanami/logger.rb
+++ b/lib/hanami/logger.rb
@@ -226,10 +226,8 @@ module Hanami
           color: :red
         )
         result << [" ", error_message, NEW_LINE].join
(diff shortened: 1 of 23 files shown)


lib/hanami/logger.rb
```diff
@@ -214,17 +215,33 @@ def _format(hash)
...
+          [ hash[:error], hash[:message] ].compact.join(": ")
+        end
+        result << [" ", error_message, NEW_LINE].join
         if hash.key?(:backtrace)
           hash[:backtrace].each do |line|
-            result << "from #{line}#{NEW_LINE}"
+            if tty?
+              result << Hanami::Utils::ShellCode.colorize("from #{line}#{NEW_LINE}", color: :yellow)
+            else
+              result << "from #{line}#{NEW_LINE}"
```
davydovanton: Just a idea, but what about `colorise` method (sorry, wired name) for wrap this tty condition? Something like this:
```ruby
private

def colorise(text, color:)
  if tty?
    Hanami::Utils::ShellCode.colorize(text, color: color)
  else
    text
  end
end
```

And call it like this:
```ruby
colorise("from #{line}#{NEW_LINE}", color: :yellow)
colorise([ hash[:error], hash[:message] ].compact.join(": "), color: :red)
```

WDYT?

lib/hanami/logger.rb
```diff
@@ -214,17 +215,33 @@ def _format(hash)
...
           end
         end
 
         result
       end
 
+      # @since x.x.x
+      # @api private
+      def tty?
+        @logdev.dev.tty?
```
davydovanton: just a question, but why `dev`? will color work for production and test envs too?

davydovanton: @jodosha thanks for a link, I wrote this before reading your comment 👍

lib/hanami/logger.rb
```diff
@@ -214,17 +215,33 @@ def _format(hash)
 
       # @api private
       def _format_error(result, hash)
-        result << " #{hash[:error]}:" if hash.key?(:error)
-        result << " #{hash[:message]}#{NEW_LINE}"
+        error_message = if tty? Hanami::Utils::ShellCode.colorize(
```
jodosha: This should be changed to:

```ruby
error_message = if tty?
  Hanami::Utils::ShellCode.colorize(
```

Otherwise Ruby interprets `Hanami::Utils::ShellCode.colorize(...)` has an argument passed to `#tty?`. It's like:

```ruby
error_message = if tty?(Hanami::Utils::ShellCode.colorize(...))
```

This is a valid syntax, but it fails because `#tty?` doesn't accept any argument.

oana-sipos: Oh, ok, thanks for noticing!

cllns:
image

One thing that's notable is that errors are *not* colorized in a Hanami app:
image
Based on our `Logger` code in this PR, the name should be red and the other lines should be yellow. I'm guessing we need to rescue `StandardError` somewhere in `hanami`'s codebase, but I'm not sure if we want to do that, since it'll make the back-trace more complicated, for a pretty minor feature. I think we should just get rid of the code the formats the error with colorization. What do you think?

This functionality doesn't need any changes to apps, it'll just work. Since the server is running in a TTY in development, the logs will be colorized.

Also, we need to decide how we want the logic to work for passing `colorize` into the logger. I couldn't figure out how to stub `tty?` for the Logger, so I couldn't implement the logic I wanted.
I was thinking we could use tri-state logic:

- `colorize: nil`, the default, will colorize only if TTY.
- `colorize: true` will colorize no matter what, even if it's not a TTY. Based on [RSpec's experience](https://github.com/rspec/rspec-core/issues/2230#issuecomment-223808366), that's useful sometimes.
- `colorize: false` will not colorize, even if it's a TTY.

@hanami/core @jodosha

cllns:
(I just pushed this branch to `hanami/utils:coloured_output`, FYI. This PR is still against Oana's branch though, since I can't change that :/

cllns:
@oana-sipos and I extracted `Colorizer` and `NullColorizer` today.

1. We didn't include the brackets in the colorizing. We had mentioned we would, and we could still do that.

2. We didn't colorize `application_name`, since this comes from outside the four arguments passed to `Logger#format_message`. It's instead injected directly into the Formatter. Not really sure how to handle that. We could have the `Colorizer` return 5 things, in a Hash, instead of an array?

3. We weren't sure about the API for `Colorizer` and `NullColorizer`. It's a little verbose and I'm not 100% sure it's the best. Maybe we could do blocks instead of methods? Maybe we could have the colors be methods rather than a Hash?

@jodosha @hanami/core

cllns:
I had a thought the other day. Isn't colorizing a responsibility of the Formatter? Rather than having the `Logger` compose the `Colorizer` and the `Formatter`, couldn't we have the `Logger` compose `Formatter`, which can include a `Colorizer`. 

If people want to provide their own `Colorizer`, it's not unreasonable to expect them to supply their own `Formatter` in the process. In fact, I _imagine_ most people who want a custom `Colorizer` will also want a custom `Formatter`, too.

cllns:
I just changed it so the `Colorizer` is injected into the `Formatter`. I think it's the right decision. People can still inject their own `Colorizer` without a `Formatter`, and they can also do the opposite.

I still haven't colorized with the brackets, because we don't need to. We _can_ do it now, by bracket-izing those parameters before joining them in the front-matter.

I'm not really sure what to do with `progname`. We don't display it by default, but it seems like we probably should?

cllns:
I just renamed `Colorizer` to `DefaultColorizer`. Let me know what you think of the recent changes @jodosha @oana-sipos :)

oana-sipos:
Hey, @cllns,

Thanks for taking care of this and for the after-thoughts. Point by point, to your questions:

1) `DefaultColorizer` sounds good.
2) Following your logic on who should be responsible for colorizing, it makes sense to be injected in the `Formatter`
3) If I had to vote, I'd probably vote for colorizing the brackets too
4) Do we really want to colorize `application_name`? Not sure it's a must.

My 2 cents.

oana-sipos:
Hey, @cllns,

Thanks for taking care of this and for the after-thoughts. Point by point, to your questions:

1) `DefaultColorizer` sounds good.
2) Following your logic on who should be responsible for colorizing, it makes sense to be injected in the `Formatter`
3) If I had to vote, I'd probably vote for colorizing the brackets too
4) Do we really want to colorize `application_name`? Not sure it's a must.

My 2 cents.

oana-sipos:
Thank you @jodosha for the review and @cllns for driving these changes!

PR #253, opened 2018-02-18
Drone CI

jodosha:
This is a experiment, please don't merge or review yet.


davydovanton:
Could you explain why drone? Maybe we can try circleCI too? WDYT?

Issue #252, opened 2018-02-16
Add #tap to Hanami::Utils::BasicObject


cllns:
I tried to simulate an error on a `Hanami::Interactor::Result` (which is a `Hanami::Utils::BasicObject`):

```ruby
Hanami::Interactor::Result.new.tap do |result|
  result.add_error("DB problem, sorry!")
end
```
and I got: ` NoMethodError: undefined method 'tap' for #<Hanami::Interactor::Result:0x00007fc90839f9a8>`

The implementation for `#tap` on `Hanami::Utils::BasicObject` should be as simple as:
```ruby
def tap
  yield self
  self
end
```
(From this blog post: http://seejohncode.com/2012/01/02/ruby-tap-that/).

I know we want to keep `Hanami::Utils::BasicObject` as minimal as possible, but this is a method that'd be useful to have. If we can't implement it on `Hanami::Utils::BasicObject`, then maybe we could add it to `Hanami::Interactor::Result` instead.

Thoughts @hanami/core ?

AlfonsoUceda:
I've never understood tap, it has the same lines than the version assigning a variable and with an extra indentation, but that's my personal opinion :P

I think we could add it, it is a very easy method.

davydovanton:
absolutely agree with @AlfonsoUceda (with `tap` and adding method :D )

davydovanton:
@cllns WDYT if we made it as `good for first PR` issue?

cllns:
@davydovanton Good idea, that's `easy` right? I just added that. (I feel like that label should be [`simple` instead of `easy`](https://www.youtube.com/watch?v=rI8tNMsozo0) 😉)

cllns:
I wonder if the better move here is to make `Hanami::Interactor::Result` _not_ inherit from `Hanami::Utils::Basic` object, and leave it as the default (`::Object`)? 

There are many other methods people could want to send to an Interactor's result:
https://ruby-doc.org/core-2.4.1/Object.html

It's not an object that's created many times, so I don't think the performance hit of using `::Object` vs `Hanami::Utils::BasicObject` will be noticeable.

sashashakun:
Hi! I am pretty new to ruby, but want to fix this, is it ok? Already checked out [contributing guide](https://github.com/hanami/utils/blob/master/CONTRIBUTING.md). Thanks!

jodosha:
@cllns Please have a look at: https://github.com/hanami/utils/pull/254#issuecomment-368882085

[...]

PR #254, opened 2018-02-23
Make Hanami::Interactor::Result inherit from ::Object

sfate:
Simply inherit `BasicObject` from stdlib's `Object` instead of picking methods from it.

Closes: https://github.com/hanami/utils/issues/252

### DIFF
--- a/lib/hanami/interactor.rb
+++ b/lib/hanami/interactor.rb
@@ -10,7 +10,7 @@ module Hanami
     # Result of an operation
     #
     # @since 0.3.5
-    class Result < Utils::BasicObject
+    class Result
       # Concrete methods
       #
       # @since 0.3.5
@@ -128,12 +128,6 @@ module Hanami
       def respond_to_missing?(method_name, _include_all)
         method_name = method_name.to_sym
         METHODS[method_name] || @payload.key?(method_name)
-      end
-
-      # @since 0.3.5
-      # @api private
-      def __inspect
-        " @success=#{@success} @payload=#{@payload.inspect}"
       end
     end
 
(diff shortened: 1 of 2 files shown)


mereghost:
Hello @sfate thanks for your contribution. From what I remember, `Hanami::Utils::BasicObject` was meant to be a slightly better `BasicObject` to reduce cases of method overriding when we return `Interactor::Result` or similar objects.

I think @AlfonsoUceda or @jodosha might have more info on this tho.

sfate:
@mereghost Agree, but at the same time there only few methods overloaded.

cllns:
Thanks for this @sfate! I meant that we should have `Hanami::Interactor::Result` (not `Hanami::Utils::BasicObject`) inherit from `::Object`. Does that make sense?

`BasicObject` is supposed to be smaller than `::Object`, so we don't want to inherit its behavior and code.

sfate:
@cllns I see, thanks. This makes more sense, right 😃 

Added changes for `Hanami::Interactor::Result` and rolled back `Hanami::Utils::BasicObject`.
/cc @mereghost

jodosha:
The reason why I picked `BasicObject` is to have the bare minimum set of methods to not overlap with any potential assigned value.

In a practical example: with `Object` as super class you can't assign a value like `select`, because it clashes with `Object#select` (from Ruby's `Kernel`). This doesn't affect any object that inherits from `BasicObject`.

So `result.select` a dynamic method from the payload, overrides `Object#select`.

---

Fast forwarding to today, I think we should change how results are exposing the payload to the outside: from dynamic methods (`result.select`) to Hash like syntax (`result[:select]`). This would remove any chance of collision between Ruby default objects and payload keys.

At that point we can safely inherit from `Object`.

That given said, we can't merge this PR or move in the direction described above because these are both breaking changes that we can ship only with 2.x series.

Issue #250, opened 2018-01-25
Proposal: inject tags to logger instance


davydovanton:
## Why?

We have long logger file with 10k lines. And we want to check what happened for specific request/job execution/etc. 

## How?
We can create simple method `#inject` (we can change name) and use it in any places:

```ruby

Hanami.logger.inject({ ... }) # => new instance of logger

new_logger = Hanami.logger.inject(uuid: uuid)

Hanami.logger.info 'hello' # => 'hello'
new_logger.info 'hello' # => 'uuid="....." hello'
```

For example, we can inject uuid into logger and use it in in rack:

```ruby

class LoggerMiddleware
  def call(env)
    env['logger'] = Hanami.logger.inject(uuid: Secure.uuid)
    [...]
  end
end
```

@hanami/core WDYT?

jodosha:
I like the idea.

AlfonsoUceda:
I like too, what do you think something more explicit?
```
Hanami.logger.tagged(uuid: uuid)
```

davydovanton:
@AlfonsoUceda yep, good idea 👍

davydovanton:
I'll start work on it 👍 

@AlfonsoUceda I think it'll be better to use `with_context` method, like: `Hanami.logger.with_context(uuid: uuid)`. WDYT?

AlfonsoUceda:
@davydovanton 👍

jodosha:
Not actionable right now.

PR #256, opened 2018-02-28
Implement PoC for with_context method

davydovanton:
Just a proof of concept for https://github.com/hanami/utils/issues/250

------------

It's just a idea but, what if we call `with_context` method with a block? For example, it will be awesome for wrapping all action:

```ruby
class LoggerMiddleware
  def call(env)
    logger.with_context(uuid: SequreRand.uuid) do
      @app.call(env)
    end
  end
end

class Books::Index
  def call(params)
    logger.info 'good request' # => [Hanami] [INFO] [2018-03-01 00:37:23 +0300] uuid="..." good request
    # ...
  end
end
```

Also, we can wrap workers, rake tasks, interactors or other places with specific context

------

I created simple and ugly PR just for proof idea. Also, I created a local file with this code:

```ruby
require 'hanami/logger'

class Test
  attr_reader :logger

  def initialize(logger: nil)
    @logger = logger
  end

  def call(payload = {})
    logger.debug 'from test object'
    logger.debug payload
  end
end


logger = Hanami::Logger.new
test = Test.new(logger: logger)

logger.info 'hello'
test.call(a: 1, b: 2)

logger.with_context(uuid: '123') do
  logger.info 'hello with context'
  test.call(a: 1, b: 2)

  logger.with_context(other: 'test') do
    logger.info 'hello with nested context'
    test.call(a: 1, b: 2)
  end

  logger.info 'bye with context'
  test.call(a: 1, b: 2)
end

logger.info 'bye'
test.call(a: 1, b: 2)
```

Anf got this result:
screenshot 2018-03-01 00 49 17

But now I have some problems with I need to solve:

- [ ] Tread safe code
- [ ] Specs

--------------

WDYT @hanami/core?

### DIFF
--- a/lib/hanami/logger.rb
+++ b/lib/hanami/logger.rb
@@ -165,6 +165,16 @@ module Hanami
       # @api private
       attr_writer :hash_filter
 
+      # @since x.x.x
+      # @api private
+      attr_writer :context
+
+      # @since x.x.x
+      # @api private
+      def context
+        @context ||= {}
+      end
+
       # @since 0.5.0
       # @api private
       #
@@ -173,7 +183,8 @@ module Hanami
         _format({
           app:      application_name,
           severity: severity,
-          time:     time
+          time:     time,
+          **context
         }.merge(
           _message_hash(msg)
         ))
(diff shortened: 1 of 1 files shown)


jodosha:
@davydovanton I'm closing this because there wasn't progress in two months. Feel free to tackle the problem with a different solution.

jodosha:
@davydovanton PS: please delete the branch if you don't need it anymore. Thanks.

davydovanton:
yep, no problem 👍

PR #258, opened 2018-04-05
Colorize logger severity

jodosha:
Colorize logger severity with the following scheme:

  * UNKNOWN => blue
  * DEBUG => cyan
  * INFO => magenta
  * WARN => yellow
  * ERROR, FATAL => red

colorize-logger-severity

### DIFF
--- a/lib/hanami/logger/colorizer.rb
+++ b/lib/hanami/logger/colorizer.rb
@@ -52,16 +52,24 @@ module Hanami
 
       private
 
-      # The colors defined for the four parts of the log message
-      #
-      # These can be overridden, keeping the name keys, with acceptable values
-      # being any from Hanami::Utils::ShellColor::COLORS
+      # The colors defined for the three parts of the log message
       #
       # @since 1.2.0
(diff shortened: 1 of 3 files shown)


lib/hanami/logger/colorizer.rb
```diff
@@ -75,7 +75,17 @@ def app(input)
       # @since 1.2.0
       # @api private
       def severity(input)
-        colorize(input, color: colors.fetch(:severity, nil))
+        color = case Hanami::Logger.level(input)
```
davydovanton: Maybe will it better to use hash of level colours instead case call?

jodosha: @davydovanton You're right. 👍 See: https://github.com/hanami/utils/pull/258/commits/d20979c7427fc956faf402ddb68f4d693cea1f46

oana-sipos:
Too late to the party (sorry, I am traveling today and tomorrow), but just to acknowledge the decision, sounds good 👍 💯

PR #269, opened 2018-04-13
Bump required_ruby_version

dogenpunk:
This pull request is leading up to a fix for hanami/hanami#921

- Add TargetRubyVersion to rubocop config to silence warnings about
mismatch between rubocop config and gemspec.
- Switch =~ usage to match? per rubocop in files.remove_block


AlfonsoUceda:
Can you update jruby to 9.1.16.0? I think it will fail because 9.2 will support 2.5 and 9.1 is still in 2.3.

PR #282, opened 2018-06-29
add Utils::IO::silence_stderr

GustavoCaso:
@jodosha 

Following @AlfonsoUceda advice I added `Utils::IO.silence_stderr`

Maybe we could look for a more general method like:

```ruby
def self.silence_output(output = $stderr)
  original_output = output
  output = File.open(File::NULL, "w")
  yield
ensure
  output = original_output
end
```

### DIFF
--- a/lib/hanami/utils/io.rb
+++ b/lib/hanami/utils/io.rb
@@ -32,6 +32,32 @@ module Hanami
       ensure
         $VERBOSE = old_verbose
       end
+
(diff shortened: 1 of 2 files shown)


GustavoCaso:
Looks like the failing CI errors has nothing to do with the added code. 😟

jodosha:
@GustavoCaso As per our agreement, we don't need this PR anymore. Thanks.

Issue #259, opened 2018-04-12
Deprecate `Utils::String` and `Utils::Hash` instance methods


jodosha:
- [x] Use `Hanami::Utils::Deprecation` for the purpose
  - [x] The deprecation warning should indicate the corresponding class method to use
  - [x] Handle `#singularize` and `#pluralize` w/o specifying the corresponding class method to use. As `Utils::String.singularize` and `.pluralize` will be removed in 2.0.
  - [x] Handle `method_missing` w/o specifying the corresponding class method to use
  - [x] Handle methods that belong to `::String` (from Ruby) w/o specifying the corresponding class method to use. Eg. `Hanami::Utils::String#gsub` is defined in our gem only for performance reasons.
  - [x] Check all the occurrences in all the Hanami gems (`develop` branches) and use the proper class method

Use the `develop` branch for this enhancement.

---

### Examples

This is for a method that is provided by `Utils::String`:

```
"Hanami::Utils::String#underscore is deprecated, please use Hanami::Utils::String.underscore"
```

This is for a method that is provided by `::String` (eg. `#gsub` or `#method_missing`:

```
"Hanami::Utils::String is deprecated as type, it will be turned into a module into Hanami 2.0"
```

---

Ref https://trello.com/c/dUwUfWrW
Ref https://trello.com/c/nJhd5awo

Ref https://github.com/hanami/utils/issues/267

PR #283, opened 2018-07-02
Deprecate `Utils::Hash` instance interface

jodosha:
Ref #259


mereghost:
Quick question: should we inform the users that the methods will be gone after version `x` releases?

AlfonsoUceda:
@mereghost we should but in a release note IMO.

mereghost:
@AlfonsoUceda Works for me.

Issue #260, opened 2018-04-12
Remove `Utils::String` as `Utils::Hash` types


jodosha:
- [x] Transform `Utils::Hash` from class into module
  - [x] Transform `Utils::String` from class into module
  - [x] Make `Utils::PathPrefix` to implicitly depend on `Object`
  - ~~Extract `Utils::String` deprecated instance methods into private class~~
  - ~~Make `Utils::PathPrefix` to depend on previous point private class~~

Use `unstable` branch for this change.

---

Ref https://github.com/hanami/utils/issues/259
Ref https://trello.com/c/dUwUfWrW

jodosha:
Implemented by #390

PR #290, opened 2018-08-05
Convert hash and string in modules

AlfonsoUceda:
This PR implements #260


AlfonsoUceda:
@jodosha yes, let me check it and I'll come back to you with the answer.

AlfonsoUceda:
Just for the records, I've search `PathPrefix` in all Hanami organisation using this query https://github.com/search?p=2&q=org%3Ahanami+Utils%3A%3APathPrefix&type=Code and I've found that we are using it in the following 3 gems:
Assets: https://github.com/hanami/assets/commits/check-path-prefix
Router: https://github.com/hanami/router/commits/check-path-prefix 
Hanami: https://github.com/hanami/hanami/commits/check-path-prefix

After CI run, I'll create a new comment with the result.

AlfonsoUceda:
@jodosha it seems I need to add `#to_sym` method because builds fail, I'm going to add it with an spec and check again.

Issue #264, opened 2018-04-12
Make `Utils::ClassAttribute` to use `Concurrent::Map`


jodosha:
Replace `::Set` usage with `Concurrent::Map`.
Please use `unstable` branch of this repository.

jodosha:
Implemented by #292

PR #292, opened 2018-08-05
Make `Utils::ClassAttribute` to use `Concurrent::Map`

AlfonsoUceda:
Thir PR implements #264

### DIFF
--- a/lib/hanami/utils/class_attribute.rb
+++ b/lib/hanami/utils/class_attribute.rb
@@ -1,7 +1,3 @@
-require 'set'
-require 'hanami/utils/duplicable'
-require 'concurrent/map'
-
 module Hanami
   module Utils
     # Inheritable class level variable accessors.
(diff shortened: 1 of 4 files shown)


lib/hanami/utils/class_attribute.rb
```diff
@@ -70,19 +71,21 @@ def class_attribute(*attributes)
             attr_accessor(*attributes)
           end
 
-          class_attributes.merge(attributes)
+          attributes.each do |attr|
+            class_attributes.put_if_absent(attr, attr)
```
jodosha: this is setting in the map both as key and value the attr name.

lib/hanami/utils/class_attribute.rb
```diff
@@ -93,7 +96,7 @@ def inherited(subclass)
         # Class accessor for class attributes.
         # @api private
         def class_attributes
-          @class_attributes ||= Set.new
+          @class_attributes ||= Concurrent::Map.new
```
jodosha: This is still not thread safe.

AlfonsoUceda: I was wondering if I had to move this attribute to other place like we do here https://github.com/hanami/utils/blob/master/lib/hanami/utils/string.rb#L77

Issue #298, opened 2018-10-03
Feature proposal: Hanami::Utils::String.squish et al


jdickey:
## Signature

```ruby
Hanami::Utils::String.squish(input_str)

Hanami::Utils::String#squish
```

## Description

`Hanami::Utils::String.squish` (and its instance-method analogue,  `String#squish`), returns the string, first removing all white-space on both ends of the string, and then changing remaining consecutive white-space groups into one space each.

## Examples

```ruby
  pp str
  # => '<span class="foo">           bar\n              Foo        the\n\tbaz</span>'
  pp Hanami::Utils::String.squish(str)
  # => '<span class="foo">bar Foo the baz</span>'
```

and

```ruby
  s2 = Hanami::Utils::String.new(str)
  # => 'Guillermo            Barranquilla -- the governor of the province --     was in the capital for consultations...and was not expected to return.'
  s2.squish
  # => 'Guillermo Barranquilla -- the governor of the province -- was in the capital for consultations...and was not expected to return.'
```

Note that all consecutive white-space is compressed to a single space, while all _other_ repeated characters are unaffected.

This should take care of the *near totality* of the inadvertent bugs introduced by cavalier use of `String#squeeze` and `String#squeeze!`.

## Prior Art

Ruby on Rails' ActiveSupport library has had a monkey-patch [`String#squish` method](https://github.com/rails/rails/commit/4ff26f51f61f9bf7d0bb34557e72c1280ee4f859) since circa Rails 2.1. It was [proposed](https://bugs.ruby-lang.org/issues/11735) for inclusion into Ruby core in November, 2015, and [shot down](https://bugs.ruby-lang.org/issues/11735#note-13) (on apparent KISS and NIH grounds)  the following June.

AMHOL:
Do you find this useful outside of heredocs? I generally only use it there and find the following syntax much more aesthetically pleasing:

```ruby
str = <<-SQL.strip.gsub(/\s+/, ' ')
  SELECT * FROM "users"
  WHERE "users"."email" = ?
SQL
```

When compared to assigning then calling `Hanami::Utils::String.squish(str)`

jdickey:
I've encountered several bugs in my own code (and specs) where I call `::String#squeeze` expecting it to "clean up" white-space while forgetting that it also removes repeated *non-white-space* characters as well.

```ruby

  s = "     Guillermo         Barranquilla\n"
  # ...
  other_s = s.squeeze
  # => " Guilermo Baranquila\n"
```
[...]

PR #301, opened 2018-10-03
Add Hanami::Utils::String.squish and #squish methods

jdickey:
As proposed in Issue #298. 

The patch previously submitted as #300 is incorporated; without that, `rspec` cannot run individual spec files. (Without #300, `bundle exec rspec spec/unit/hanami/utils/string_spec.rb` will fail but `bundle exec rake` will pass due to a missing `require`.)

### DIFF
--- a/lib/hanami/utils.rb
+++ b/lib/hanami/utils.rb
@@ -1,5 +1,8 @@
 # Hanami - The web, with simplicity
 #
+
+require 'pathname'
+
 # @since 0.1.0
 module Hanami
   require 'hanami/utils/version'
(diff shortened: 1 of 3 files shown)


lib/hanami/utils.rb
```diff
@@ -1,5 +1,8 @@
 # Hanami - The web, with simplicity
 #
+
+require 'pathname'
```
davydovanton: sorry, I can't understand this line. Why we need to require `pathname` in base utils file? can we do it only in specifics files?

davydovanton: got it, thanks for explanation!

jodosha:
@jdickey Thanks for this PR, but I'm closing it. Please read: https://github.com/hanami/utils/issues/298#issuecomment-451174454

PR #312, opened 2018-11-16
Introduce Callbacks::Chain#dup

jodosha:


jodosha:
It's an old branch that I had with this fix. You're too fast to review 😛 . That's why there wasn't even a description to the PR. 😉

PR #315, opened 2019-01-02
Ruby 2.6.0

jodosha:


jodosha:
Not including yet JRuby 9.2.1.0 because it doesn't work properly with RSpec. When there is an assertion failure, JRuby exits the process, and doesn't print the post-failure RSpec message.

https://travis-ci.org/hanami/utils/jobs/474382234

PR #317, opened 2019-01-20
Add instance_of?, is_a?, kind_of? To BasicObject

funk-yourself:
The lack of Hanami::Interactor::Result#is_a? causes my test suite to error in such examples:

```ruby
it { is_expected.to satisfy(&:failure?) }
```

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,0 +1,9 @@
+kind: pipeline
+name: default
+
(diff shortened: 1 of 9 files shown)


jodosha:
@funk-yourself Thanks again for this PR. Sorry for taking so long.

PR #321, opened 2019-01-28
Utils::String module

jodosha:
## Changes

  * Remove instance level interface for `Utils::String`
  * Turn `Utils::String` from class to module
  * Make `Utils::PathPrefix` to not inherit from `Utils::String` anymore

---

Ref https://github.com/hanami/utils/issues/260


jodosha:
Closing this as the taken path is complex.

PR #333, opened 2019-05-25
Utils::String#underscore handles dots

IvanShamatov:
I first stuck to this problem when decided to generate new hanami project.
```
hanami new project.name
```
this command will create `module Project.name`, which is definitely incorrect and fails. Though, it's quite common to generate projects with the first level domain like `hanamirb.org`.

I've decided that this logic best suit exactly inside `underscore` and doesn't require a separate method for that. Especially in case, it already has whitespace- and dash- filtering.

### DIFF
--- a/lib/hanami/utils/string.rb
+++ b/lib/hanami/utils/string.rb
@@ -253,7 +253,7 @@ module Hanami
         string.gsub!(NAMESPACE_SEPARATOR, UNDERSCORE_SEPARATOR)
         string.gsub!(/([A-Z\d]+)([A-Z][a-z])/, UNDERSCORE_DIVISION_TARGET)
         string.gsub!(/([a-z\d])([A-Z])/, UNDERSCORE_DIVISION_TARGET)
-        string.gsub!(/[[:space:]]|\-/, UNDERSCORE_DIVISION_TARGET)
+        string.gsub!(/[[:space:]]|\-|\./, UNDERSCORE_DIVISION_TARGET)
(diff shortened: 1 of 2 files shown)


IvanShamatov:
Though, it looks like it is a problem of classify

jodosha:
@IvanShamatov Thanks for this PR, but I don't think it's a good idea to merge it for two reasons:

  1. It's a breaking change.
  1. If the naming of a new Hanami app doesn't work as expected, then we should fix the problem in `hanami`, not here. Assuming we're using `Utils::String.underscore` in the implementation of `hanami new`, the day we'll change the implementation with let's say `Dry::Inflector#underscore` (just an example), we could introduce a regression in that CLI command.

Thanks for understanding. 👍

IvanShamatov:
Sure, no problem, it just made me think, that there are no definition for `underscore`. For example, why spaces become underscores and dots don't? I thought it was lack of implementation. Thats why I placed it in utils, not Hanami repo. 
Im not arguing, you are the boss ;)
Cheers

jodosha:
@IvanShamatov True, it isn't set in stone, but it's more an implicit result taken from `activesupport` legacy and eventually spread across the Ruby ecosystem.

The key points are:

  1. it's a breaking change that we can't introduce now
  2. adding this new behavior to `.underscore` doesn't guarantee that `hanami` won't have the same bug again, if the implementation there will no longer use `.underscore`, but a different method to transform the name.

PR #340, opened 2019-08-07
Apply singular for snake_cased words

### LINKED ISSUE #339: Singularize get's confused with snake_case words
I was trying hanami and created a resource called `:control_exercises`. I wasn't able to use the path helper so I listed my routes and saw this:

```
                root GET, HEAD  /                              Web::Controllers::Home::Index
     control_exerces GET, HEAD  /control_exercises             Web::Controllers::ControlExercises::Index
 new_control_exercis GET, HEAD  /control_exercises/new         Web::Controllers::ControlExercises::New
     control_exerces POST       /control_exercises             Web::Controllers::ControlExercises::Create
```

Digging deeper I came into the utils repo and found that the problem is that when passed through inflector singularize, the special word `exercise` doesn't get picked because of the `control_` prefix.

https://github.com/hanami/utils/blob/2cc692fb38be8f8233a403ac72ad465a4f906f32/lib/hanami/utils/inflector.rb#L428

mauromorales:
I'm super new to hanami and all its ecosystem, this PR is my naive approach at how I'd solve the problem of handling snake cased words which are part of the special singulars/plurals e.g. `control_exercises`.

Please give me some feedback and I'll be happy to apply to the best of my knowledge

fixes #339

### DIFF
--- a/lib/hanami/utils/inflector.rb
+++ b/lib/hanami/utils/inflector.rb
@@ -27,17 +27,31 @@ module Hanami
         # @since 0.6.0
         # @api private
         def ===(other)
-          key = other.downcase
+          key = if other.downcase =~ /_(\w*)\z/
+            $1
+          else
+            other.downcase
+          end
           @rules.key?(key) || @rules.value?(key)
         end
 
         # @since 0.6.0
         # @api private
         def apply(string)
-          key    = string.downcase
+          key = if string.downcase =~ /_(\w*)\z/
+            $1
+          else
+            string.downcase
+          end
           result = @rules[key] || @rules.rassoc(key).last
 
-          string[0] + result[1..-1]
+          prefix = if key == string.downcase
+            string[0]
+          else
+            string[0..string.index(key)]
+          end
+
+          prefix + result[1..-1]
         end
       end
 
--- a/lib/hanami/utils/inflector.rb
+++ b/lib/hanami/utils/inflector.rb
@@ -27,31 +27,31 @@ module Hanami
         # @since 0.6.0
         # @api private
         def ===(other)
-          key = if other.downcase =~ /_(\w*)\z/
-            $1
-          else
-            other.downcase
-          end
+          key = last_alpha(other)
           @rules.key?(key) || @rules.value?(key)
         end
 
         # @since 0.6.0
         # @api private
         def apply(string)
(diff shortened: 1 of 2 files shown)


spec/unit/hanami/utils/string_spec.rb
```diff
@@ -352,6 +352,11 @@
       expect(result).to eq(@singular)
     end
 
+    it 'singularizes a composed string', silence_deprecations: true do
```
jodosha:
```suggestion
    it 'singularizes a snake_case string', silence_deprecations: true do
```

lib/hanami/utils/inflector.rb
```diff
@@ -27,17 +27,31 @@ def add(key, value)
         # @since 0.6.0
         # @api private
         def ===(other)
-          key = other.downcase
+          key = if other.downcase =~ /_(\w*)\z/
```
jodosha: You may want to change `\w` (A word character `[a-zA-Z0-9_]`) with `[[:alpha:]]` (Alphabetic character), to match chars beyond ASCII set.

lib/hanami/utils/inflector.rb
```diff
@@ -27,17 +27,31 @@ def add(key, value)
...
+            other.downcase
+          end
           @rules.key?(key) || @rules.value?(key)
         end
 
         # @since 0.6.0
         # @api private
         def apply(string)
-          key    = string.downcase
+          key = if string.downcase =~ /_(\w*)\z/
```
jodosha: Please consider to extract a private method for this logic, as it's exactly the same used for `#===`

lib/hanami/utils/inflector.rb
```diff
@@ -27,17 +27,31 @@ def add(key, value)
...
+          prefix = if key == string.downcase
+                     string[0]
+                   else
+                     string[0..string.index(key)]
+                   end
+
+          prefix + result[1..-1]
+        end
+
+        def last_alpha(string)
```
mauromorales: @jodosha I see you use `# @api private` what is that about? I wanted to put `last_alpha` under the `private` access control but I thought I'd ask about that first.

mauromorales:
@jodosha all changes you requested should be done by now, please have a look and let me know if you have any feedback, thanks

PR #341, opened 2019-08-23
Run specs in TruffleRuby in Travis CI

deepj:
Since TruffleRuby 19.2.0 all Hanami::Utils' specs pass 🚀

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,12 +12,14 @@ rvm:
   - jruby-9.1.9.0
   - ruby-head
   - jruby-head
+  - truffleruby
 
 matrix:
   allow_failures:
     - rvm: ruby-head
     - rvm: jruby-head
     - rvm: jruby-9.1.9.0
+    - rvm: truffleruby
 
 notifications:
   webhooks:


jodosha:
@davydovanton @deepj Thanks for taking care of this. Because we're experimenting DroneCI (see https://ci.hanamirb.org), we should aim for CI parity of Ruby versions.

@deepj Could you please have a look at https://github.com/hanami/docker-ci-images and build a Truffle Docker image based on the current ones that we have on that repo? Thanks in advance.

deepj:
@jodosha Sorry for late reply. I'll try

.travis.yml
```diff
@@ -12,12 +12,14 @@ rvm:
...
   - ruby-head
   - jruby-head
+  - truffleruby
 
 matrix:
   allow_failures:
     - rvm: ruby-head
     - rvm: jruby-head
     - rvm: jruby-9.1.9.0
+    - rvm: truffleruby
```
eregon: It should not be in `allow_failures`, otherwise nobody will notice it fails typically.
This might have shown that adding RuboCop would break the TruffleRuby build for https://github.com/hanami/utils/pull/344

PR #342, opened 2019-09-23
Fix error of logger's filter when input have closed Tempfile

lfidnl:
Hello.

I faced up with issue during upload a file to hanami app. Application works fine, but logger raise `IOError: closed stream`. This happens because logger's filter tries to navigate through closed Tempfile.


lfidnl:
@jodosha hello! I've fixed it

PR #344, opened 2019-09-25
Revert "Run specs in TruffleRuby in Travis CI"

jodosha:
Reverts hanami/utils#341

TruffleRuby build keeps 16 minutes to complete and it's constantly failing. See an example: https://travis-ci.org/hanami/utils/jobs/589308192

This isn't a definitive NO to TruffleRuby. But for now it's just a waste of time and resources to keep it as it is in CI build.

/cc @davydovanton @deepj

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,14 +12,12 @@ rvm:
   - jruby-9.1.9.0
   - ruby-head
   - jruby-head
-  - truffleruby
 
 matrix:
   allow_failures:
     - rvm: ruby-head
     - rvm: jruby-head
     - rvm: jruby-9.1.9.0
-    - rvm: truffleruby
 
 notifications:
   webhooks:


deepj:
@jodosha It seems the only problem lies in Rubocop. Can it be disabled for TruffleRuby. I’m just pinging @eregon if he’s interesting in this

eregon:
It would be interesting to see if specs pass.
I'm a bit surprised RuboCop fails, as it usually works fine, but indeed it's not efficient to run RuboCop in every CI job.

One way to make the build faster would be to not do `gem update --system`. That makes every job slower, and more prone to fail due to unrelated changes. It shouldn't be necessary anymore for Ruby 2.3+ anyway.

eregon:
@deepj Could you try to make a PR that skips RuboCop for the TruffleRuby job, still runs the specs, no `allow_failures` (to notice when it starts failing) and no `gem update --system`? I'd do it myself but I'm quite busy working on TruffleRuby itself currently.

deepj:
@eregon in fact, there is some issue with `MultiJson` gem with the recent TruffleRuby. At least in the time when I sent a PR adding running TruffleRuby on Travis CI, everything was green.

deepj:
I did make these changes https://github.com/deepj/utils/commit/dff08b0729ff94738c46f2566be1f05b180519e9

@jodosha any thoughs if this can be acceptable?

eregon:
@deepj 
> @eregon in fact, there is some issue with `MultiJson` gem with the recent TruffleRuby. At least in the time when I sent a PR adding running TruffleRuby on Travis CI, everything was green.

Could you show me the output?

Could you open a PR? Then it's easier to see the CI and track what needs to be done.

jodosha:
@eregon Thanks for joining this conversation. I really appreciate that! 💚 

> I'm a bit surprised RuboCop fails, as it usually works fine, but indeed it's not efficient to run RuboCop in every CI job.

The point is: we use Rubocop. Even if we run it only once a month against Truffle, it will still cause a CI failure. Should we give up running Rubocop against Truffle? If you know that is not gonna work, I'm fine with that.

> One way to make the build faster would be to not do gem update --system. That makes every job slower, and more prone to fail due to unrelated changes. It shouldn't be necessary anymore for Ruby 2.3+ anyway.

That is there for historical reasons :tm: 😜.

That was necessary because Travis CI messing with their Linux images, or there was an outdate Rubygems version installed there that should've been update to prevent the build to error. Reaching their support asked to run `gem update --system` to solve a problem as one time shot, but then we forgot to remove that from `.travis.yml`.

Are you suggesting remove that command to speed up the build?

---

Thanks for all the inputs.

jodosha:
@deepj I endorse @eregon suggestion to open another PR so we can see what you see via CI builds. Thanks.

eregon:
> The point is: we use Rubocop. Even if we run it only once a month against Truffle, it will still cause a CI failure. Should we give up running Rubocop against Truffle? If you know that is not gonna work, I'm fine with that.

Absolutely. I would just suggest running RuboCop for only 1 of the matrix's jobs instead of all, but just skipping it for TruffleRuby would already be enough.
Even if it worked, it would be redundant with RuboCop running on MRI as part of another job of the matrix, and make the TruffleRuby job slower for no added value.

> Are you suggesting remove that command to speed up the build?

Yes indeed, it was needed for Bundler 2.0.0 but not anymore for Bundler 2.0.1:
https://bundler.io/blog/2019/01/04/an-update-on-the-bundler-2-release.html
That unfortunately is still not clearly explained in Bundler and TravisCI docs.

PR #347, opened 2019-10-16
Store raised exception in rack.errors and rack.exception

lfidnl:
@jodosha hello agian :)
Based on our discussion in https://github.com/hanami/hanami/pull/1014 I created this PR. This common utility method can be used in action and in view.
After success merging this PR I can create 2 PRs: one for action (https://github.com/hanami/controller/blob/c12e07954d56b5df5da6e5f9dd05f48634ebc38e/lib/hanami/action/throwable.rb#L160) and one for view (https://github.com/hanami/hanami/blob/master/lib/hanami/rendering_policy.rb#L60)

### DIFF
--- a/lib/hanami/utils/rack_error.rb
+++ b/lib/hanami/utils/rack_error.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+module Hanami
+  module Utils
+    # This module provides method for saving exceptions
+    # to rack-env:
+    # <tt>rack.errors</tt> - for common Rack SPEC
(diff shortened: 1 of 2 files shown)


jodosha:
@lfidnl Thanks for this quick iteration. I don't think that this is the right place for this utility. Given this provides an utility for Rack, it should go in `hanami-controller`.

The other component that will use this utility is `hanami` which depends on `hanami-controller`. There is no reason to involve `hanami-utils`.

Thanks. 💚

PR #349, opened 2019-10-18
`Utils::Files.append` ensure to work without newline

### LINKED ISSUE #348: Utils::Files.append doesn't add properly contents when file doesn't end with a newline
If routes.rb doesn't end with a newline, generate action adds a line with syntax error.

This came up while following the new user guide. Although it should be second nature to have newlines at the end of your files, it's certainly not a requirement in Ruby. While following the guide and copy pasting code, I ended up with a broken routes file after running a generator.

To reproduce:

Edit your routes.rb file to not have a newline, something like 

```
root to: 'home#index'
```

Run a generator:

```
hanami g action web sunshine#index
```

The resulting file is 

```
root to: 'home#index'get '/tires', to: 'sunshine#index'

```

It would be nice if the template generator could detect that the last char.

I tracked the file change back to https://github.com/hanami/utils/blob/master/lib/hanami/utils/files.rb#L162 -- so perhaps it would make more sense to create the issue there instead?

jodosha:
`Utils::Files` ensure to properly append content when a file doesn't terminate with a newline

Closes #348

### DIFF
--- a/spec/unit/hanami/utils/files_spec.rb
+++ b/spec/unit/hanami/utils/files_spec.rb
@@ -246,6 +246,19 @@ RSpec.describe Hanami::Utils::Files do
       expect(path).to have_content(expected)
     end
 
+    # https://github.com/hanami/utils/issues/348
+    it "adds a line at the top of a file that doesn't end with a newline" do
+      path = root.join("unshift_missing_newline.rb")
+      content = "get '/tires', to: 'sunshine#index'"
+
+      described_class.write(path, content)
+      described_class.unshift(path, "root to: 'home#index'")
+
+      expected = "root to: 'home#index'\nget '/tires', to: 'sunshine#index'"
+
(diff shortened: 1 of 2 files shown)


spec/unit/hanami/utils/files_spec.rb
```diff
@@ -246,6 +246,19 @@ class Unshift
       expect(path).to have_content(expected)
     end
 
+    # https://github.com/hanami/utils/issues/348
+    it "adds a line at the top of a file that doesn't end with a newline" do
```
ttilberg: This should be "add a line at the *bottom* of a file..."

jodosha: I've added a spec for `.unshift` as well, which adds a line at the _top_ of the file. The spec for `.append` is at line 296. 😉

PR #350, opened 2019-10-25
API docs: unify documentation approach

dimaboyko:
I have noticed, different methods are using different approach to describe. The purpose ("to do smth") vs. the action ("does smth").
With this PR I tried to make it more consistent across the files.

### DIFF
--- a/lib/hanami/utils/hash.rb
+++ b/lib/hanami/utils/hash.rb
@@ -121,15 +121,15 @@ module Hanami
         end
       end
 
-      # Shallow duplicates hash values
+      # Deep duplicates hash values
       #
-      # The output of this function is a shallow duplicate of the input.
+      # The output of this function is a deep duplicate of the input.
       # Any further modification on the input, won't be reflected on the output
       # and viceversa.
       #
       # @param input [::Hash] the input
       #
-      # @return [::Hash] the shallow duplicate of input
+      # @return [::Hash] the deep duplicate of input
       #
       # @since 1.0.1
       #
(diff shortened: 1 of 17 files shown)


lib/hanami/utils/hash.rb
```diff
@@ -121,7 +121,7 @@ def self.deep_stringify(input) # rubocop:disable Metrics/MethodLength
         end
       end
 
-      # Deep duplicate hash values
+      # Shallow duplicates hash values
```
dimaboyko: Honestly, not sure about that one. description and naming seem a bit contradicting

dimaboyko:
waiting for that one  -> https://github.com/colszowka/simplecov/pull/754

lib/hanami/utils/hash.rb
jodosha: > description and naming seem a bit contradicting

The semantic of this method is: duplicate (as in `#dup`) all the nested values of the input hash.

Trivial example:

```ruby
hash = { a: [1, 2, 3] }
hash.fetch(:a).object_id
# => 70340898593420
duplicated = Hanami::Utils::Hash.deep_dup(hash)
# => {:a=>[1, 2, 3]}
duplicated.fetch(:a).object_id
# => 70340898530980
```

---

What's the best way to express it?

dimaboyko: So it behaves like classic deep_dup. Then I think it would make sense to reflect that in the name as well as in description, which now says: 

> The output of this function is a shallow duplicate of the input.

will push such change.

PR #379, opened 2020-08-31
Consider keyword arguments in the interactor.

ippachi:
```
require 'hanami/interactor'

class A
  include Hanami::Interactor
  def call(test, a:, b:)
    puts test, a, b
  end
end

A.new.call({ test: 1 }, a: 2, b: 3)
```

This code output warning about the keyword argument of ruby 2.7.
So I declared the keyword argument explicitly.

### DIFF
--- a/lib/hanami/interactor.rb
+++ b/lib/hanami/interactor.rb
@@ -355,7 +355,15 @@ module Hanami
       #   Signup.new.call # => NoMethodError
       def call(*args)
         @__result = ::Hanami::Interactor::Result.new
-        _call(*args) { super }
+        _call(*args) do
+          if method(:call).super_method.parameters.last[0] == :keyrest &&
+             args.last.respond_to?(:to_hash)
+            kwargs = args.pop
+            super(*args, **kwargs)
+          else
+            super(*args)
+          end
+        end
       end
 
       private
--- a/spec/unit/hanami/interactor_spec.rb
+++ b/spec/unit/hanami/interactor_spec.rb
@@ -294,13 +294,13 @@ class CreateUser
   expose :user
 
   def call(**params)
-    build_user(params)
+    build_user(**params)
     persist
   end
 
   private
 
-  def build_user(params)
+  def build_user(**params)
     @user = User.new(params)
   end
 


jodosha (changes requested):
@ippachi Thanks for this PR, but..

It doesn't fix all the warnings:

```
/Users/luca/Code/hanami/utils/spec/unit/hanami/interactor_spec.rb:522: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/Users/luca/Code/hanami/utils/lib/hanami/interactor.rb:354: warning: The called method `call' is defined here
F/Users/luca/Code/hanami/utils/spec/unit/hanami/interactor_spec.rb:295: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/Users/luca/Code/hanami/utils/spec/unit/hanami/interactor_spec.rb:318: warning: The called method `build_user' is defined here
```

...and it causes a spec failure:

```

Failures:

  1) Hanami::Interactor new interface #call handles args with to_hash method
     Failure/Error: expect(result.args).to eql(Array[user])

       expected: [#<User:0x00007f965d27c950 @attributes={:name=>"Luca"}>]
            got: [{:name=>"Luca"}]

       (compared using eql?)

       Diff:
       @@ -1 +1 @@
       -[#<User:0x00007f965d27c950 @attributes={:name=>"Luca"}>]
       +[{:name=>"Luca"}]

     # ./spec/unit/hanami/interactor_spec.rb:523:in `block (4 levels) in <top (required)>'
```

Please consider to rebase with `master` and then fix the code. Thanks. 😃

ippachi:
Sorry incompleted PR.
I'll fix it! 😄

ippachi:
I fixed.
This is not clean code so I think there is better way.
But I have no idea. 😢 

@jodosha

jodosha (changes requested):
@ippachi This is a hacky solution. 😞

jodosha:
With v1.3.7 and the upcoming v1.3.8 we'll have this to work well with both args and kwargs with Ruby 2.7 and 3.0.

PR #385, opened 2020-09-26
Change archived gem "Transproc" for its successor "Dry::Transformer"

emirdaponte:
Hi, I was looking at the **Utils::Hash** class in the `unstable` branch and noticed that they are using the gem "transport" that was archived and is now "dry-transformer".

I checked the trello (Hanami 2.0) just in case and did not see any tasks planned for this change

I changed the gem and made some small changes to fit the gem Dry::Transformer.


jodosha (approved):
@emirdaponte Thanks for this PR. 👍

PR #389, opened 2020-10-15
Allows Logger can filter an array of hash

khaile:
Currently, the `Logger` support filter only for a single hash. In case the input sent to the logger is an array of hash, the input does not filter as expected.

Before:
- Input
```ruby
[{:password=>"foo", :user=>{:name=>"foo", :password=>"abc"}}, {:password=>"bar", :user=>{:name=>"bar", :password=>"xyz"}}]
```
- Output
```ruby
[{:password=>"foo", :user=>{:name=>"foo", :password=>"abc"}}, {:password=>"bar", :user=>{:name=>"bar", :password=>"xyz"}}]
```

After:
- Input
```ruby
[{:password=>"foo", :user=>{:name=>"foo", :password=>"abc"}}, {:password=>"bar", :user=>{:name=>"bar", :password=>"xyz"}}]
```
- Output
```ruby
[{:password=>"foo", :user=>{:name=>"foo", :password=>"[FILTERED]"}}, {:password=>"bar", :user=>{:name=>"bar", :password=>"[FILTERED]"}}
```

### DIFF
--- a/spec/unit/hanami/logger_spec.rb
+++ b/spec/unit/hanami/logger_spec.rb
@@ -659,6 +659,34 @@ RSpec.describe Hanami::Logger do
       expect(output).to eql(Hash["password" => "[FILTERED]", foo: Hash[password: "[FILTERED]"]])
     end
 
+    it "doesn't mutate the input hash" do
+      input = Hash["password" => "azerty", foo: Hash[password: "bar"]]
+      described_class.new(%w[password]).call(input)
+      expect(input).to eql(Hash["password" => "azerty", foo: Hash[password: "bar"]])
+    end
+
+    it "doesn't mutate the input array" do
+      input = Array[Hash["password" => "azerty", foo: Hash[password: "bar"]]]
+      described_class.new(%w[password]).call(input)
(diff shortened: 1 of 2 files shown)


spec/unit/hanami/logger_spec.rb
```diff
@@ -672,9 +700,10 @@ class TestLogger < Hanami::Logger; end
     end
 
     context "when input has Tempfile" do
-      it "filters with out errors with closed stream" do
-        tempfile = Tempfile.new("filter_test")
-        tempfile.close!
+      let(:tempfile) { Tempfile.new("filter_test") }
+      after { tempfile.close! }
```
hieuk09, line 704: @khaile this line changes the semantic of the test. The test tests the case when file is closed before running this log method, while your change make it close after the log method.

khaile: @hieuk09 thank you for clarifying. I will update my change since this line against the purpose of this PR https://github.com/hanami/utils/pull/342/files I haven't seen before 🙇‍♂️

jodosha (approved):
@khaile Thanks for this fix. LGTM. Merging 👍

PR #390, opened 2020-12-04
Transform `Utils::String` into module

jodosha:
## What?

Turn `Utils::String` from a class to a module.

## Why?

Wrapping Ruby core's `::String` into a custom object was a design mistake. As we moved to a F/OOP approach, it makes sense to have a module to provide string manipulation utilities.

## Potential Enhancements

The set of features that `Utils::String` implements overlap with `dry-inflector`, it's worth considering to get rid of `Utils::String` altogether.

But for the present, turning into a module it's a great opportunity to remove all the instances of `Utils::String` from Hanami codebase.

---

Closes https://github.com/hanami/utils/issues/260


solnic (approved):
This makes sense!

PR #392, opened 2021-02-03
Allow keyword arguments in interactor initializer for ruby 3

hieuk09:
I have an interactor like this:

```ruby
class CreateUser
  include Hanami::Interactor

  def initialize(user_repo: UserRepository.new)
    @user_repo = user_repo
  end

  def call(**params)
    user_repo.create(params)
  end
end
```
When upgrade to ruby 3, this interactor raises `ArgumentError: wrong number of arguments (given 1, expected 0)` when initialized.

### DIFF
--- a/lib/hanami/interactor.rb
+++ b/lib/hanami/interactor.rb
@@ -181,10 +181,18 @@ module Hanami
       #       # ...
       #     end
       #   end
-      def initialize(*args)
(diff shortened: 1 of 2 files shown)


jodosha (approved):
@hieuk09 Thanks for this enhancement. 👍

PR #395, opened 2022-01-20
`Hanami::Logger`: allow for to specify `File::NULL` as `stream:`

robjacoby:
If the `stream` is set to `File::NULL`, `@logdev` will be `nil`

This is due to the following logic in `Logger#initialize`
https://github.com/ruby/logger/blob/cdeddb2f6721e9003d71a7013d68733dd992b0d6/lib/logger.rb#L388-L394

The use case for this is if you wish to "throw away" any logs without
polluting stdout or a logfile. Perhaps in testing for instance where you might want `Hanami::Logger.new(stream: "/dev/null")`

### DIFF
--- a/lib/hanami/logger.rb
+++ b/lib/hanami/logger.rb
@@ -354,6 +354,8 @@ module Hanami
     # @since 1.2.0
     # @api private
     def tty?
+      return false if @logdev.nil?
+
       @logdev.dev.tty?
     end
   end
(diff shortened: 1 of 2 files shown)


jodosha (approved):
@robjacoby Thanks for this PR. LGTM 👍 

Out of curiosity, why are you setting `File::NULL`? Is it during unit tests?

robjacoby:
> Is it during unit tests?

Yep almost always during tests. I often don't want logging output polluting my test output, and so pass in a "null" logger to throw away the values. That way I have an actual object (rather than a double) to help ensure there are no weird regressions.

PR #397, opened 2022-05-06
Replace period in underscored string

tak1n:
Related to https://github.com/hanami/hanami/issues/1163


tak1n:
Should I open a separate PR against `main` as it still seems to use the same `underscore` method as well?

adam12:
The fix looks good, but I am not sure how to approach this because it is backwards breaking behaviour. Probably best to let @jodosha look at it when he has a free moment.

jodosha (changes requested):
@tak1n Please take also care of the [instance level interface](https://github.com/hanami/utils/blob/0ff863325b9c0f6a003d2125683cd74708aa7ea5/lib/hanami/utils/string.rb#L488).

tak1n:
> @tak1n Please take also care of the [instance level interface](https://github.com/hanami/utils/blob/0ff863325b9c0f6a003d2125683cd74708aa7ea5/lib/hanami/utils/string.rb#L488).

Done

jodosha (approved):
@tak1n Thank you! 👍

PR #399, opened 2022-05-16
Move `Hanami.env` and `Hanami.env?` into `Hanami::Utils`

tak1n:
Tracked by https://trello.com/c/dpt5iSkY/108-add-hanamienv-and-hanamienv-to-hanami-utils-so-we-can-rely-on-them-in-gems-like-hanami-controller-when-they-are-used-standalone

Let me know whether you would like to have those methods defined in `Hanami::Utils::Env` instead of the root `Hanami::Utils` module.

### DIFF
--- a/lib/hanami/utils.rb
+++ b/lib/hanami/utils.rb
@@ -46,7 +46,6 @@ module Hanami
     # @return [Symbol] env
     #
     # @since 2.0.0
-    # @api private
     def self.env
       ENV.fetch("HANAMI_ENV", "development").to_sym
     end
@@ -55,10 +54,9 @@ module Hanami
     #
     # @param [Array<String, Symbol>] list of environment names
     #
-    # @return [TrueClass,FalseClass]
+    # @return [Boolean]
     #
     # @since 2.0.0
-    # @api private
     def self.env?(*names)
       names.map(&:to_sym).include?(env)
     end
--- a/lib/hanami/utils.rb
+++ b/lib/hanami/utils.rb
@@ -41,6 +41,28 @@ module Hanami
       RUBY_ENGINE == HANAMI_RUBINIUS
     end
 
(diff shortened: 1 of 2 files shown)


timriley:
Thanks for taking this on, @tak1n!

Interestingly, I only last month moved the "integrate with a full Hanami app" parts of hanami-view and hanami-controller back into the Hanami gem itself, so our immediate need for the relocation of `Hanami.env` into Hanami-utils is less critical now.

So at this point we should probably step back and reconsider whether this change still makes sense. @jodosha — can you imagine scenarios where it might be helpful to have these in `hanami-utils` in addition to (or instead of) `hanami` itself?

The biggest area that comes to my mind is allowing 3rd party gems to detect whether an Hanami application is present before offering any kind of integration shims. However, I'm not sure if that's really served by `.env` and `.env?`, so my thought at this point is that it's probably better just to close this PR for now (sorry!) and keep the idea in mind if we ever get to a point where it makes sense again.

tak1n:
> The biggest area that comes to my mind is allowing 3rd party gems to detect whether an Hanami application is present before offering any kind of integration shims. However, I'm not sure if that's really served by .env and .env?, so my thought at this point is that it's probably better just to close this PR for now (sorry!) and keep the idea in mind if we ever get to a point where it makes sense again.

I see. Thx for the clarifying response, feel free to close if no longer appropriate ✌🏽

tak1n:
[Add Hanami.env and Hanami.env? to hanami-utils, so we can rely on them in gems like hanami-controller when they are used standalone](https://trello.com/c/dpt5iSkY/108-add-hanamienv-and-hanamienv-to-hanami-utils-so-we-can-rely-on-them-in-gems-like-hanami-controller-when-they-are-used-standalone)

PR #404, opened 2022-11-13
Convert to Zeitwerk for code loading

timriley:
Covert the gem to use Zeitwerk for code loading. To do this:

- Add zeitwerk to the gemspec
- Set up the Zeitwerk loader inside `lib/hanami/utils.rb` (following the same `.gem_loader`  name I've used in all the other Hanami gems I'm converting today, since it's clear and now consistent)
- Remove all now-unneeded manual requires for files inside the gem

Of all the zeitwerk conversions, this one _may_ result in some potential breakages since a cherry-picked require without a _previous_ require to `"hanami/utils"` will mean that Zeitwerk is not set up. However, since usage of hanami-utils is almost entirely in-ecosystem, we can sort out any issues in the next day or two.

### DIFF
--- a/hanami-utils.gemspec
+++ b/hanami-utils.gemspec
@@ -22,6 +22,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency "dry-transformer", "~> 0.1"
   spec.add_dependency "concurrent-ruby", "~> 1.0"
+  spec.add_dependency "zeitwerk", "~> 2.6"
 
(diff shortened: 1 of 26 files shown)


timriley:
[Zeitwerkify hanami-utils](https://trello.com/c/PZgZ4p1I/330-zeitwerkify-hanami-utils)

jodosha:
We agreed to not proceed.

PR #406, opened 2022-11-28
Make chains comparable by #== based on contents

timriley:
This is useful behavior on its own, but is also necessary to ensure callbacks from a base app-level Hanami::Action class propogate all the way through a slice-level base class to an individual action class.

This adds dry-core to the list of gem dependencies to get access to its easy-to-use `Dry::Equalizer`. This feels like a reasonable addition given that dry-core is already going to be in the bundle for any project using Hanami gems.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 Ruby core extensions and class utilities for Hanami
 
+## Unreleased
+
+### Changed
+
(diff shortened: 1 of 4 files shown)


jodosha (approved):
@timriley I'm fine with this change.
Is it a fix for a bug?

timriley:
@jodosha Yep! It fixes the bug where callback chains containing callbacks defined in a base app action class actually get cleared out instead of properly inherited when slice-based action classes inherit from the app action class.

Issue #408, opened 2023-01-17
self check in #blank?


hi-tech-jazz:
Shouldn't this be `!object` instead of `!self`? 🤔 

https://github.com/hanami/utils/blob/e8916251c4988331c7ea6fdc1c3ba09009b6d736/lib/hanami/utils/blank.rb#L44

Looks like a copy paste from the ActiveSupport monkeypatch [implementation](https://www.rubydoc.info/gems/activesupport/Object:blank%3F).

jodosha:
@hi-tech-jazz Yes, you're right.
Are you in for a contribution to fix the problem?

Thank you 🙂

jodosha:
Fixed by https://github.com/hanami/utils/pull/409

PR #409, opened 2023-01-18
reference target object instead of self in .blank?

hi-tech-jazz:
Addresses https://github.com/hanami/utils/issues/408. It looks like part of the existing `blank?` implementation was copied from from the Active Support counterpart and was not properly updated to fit the new context.


jodosha (approved):
@hi-tech-jazz Thanks for this PR. LGTM. 👍

PR #410, opened 2023-04-15
Remove Escape

timriley:
This functionality has been moved to hanami-view for 2.0 in https://github.com/hanami/view/pull/229.

I've done a [GitHub code search for "Escape."](https://github.com/search?q=org%3Ahanami%20Escape.&type=code) in the hanami org and it looks like the only main branch dependency we have on this module is in hanami-helpers’ `HtmlHelper`, which has been taken care of in https://github.com/hanami/view/pull/229 too, making this change here safe to merge for 2.0.0.


parndt:
This was also used in hanami-controller's specs: https://github.com/hanami/controller/actions/runs/4830318934/jobs/8606398791

PR #413, opened 2024-12-25
Add support for Ruby 3.4

cllns:
We still get ostruct warnings but it's only used in a couple specs so I think we should rework the specs to not use OpenStruct... or we could just add it to the Gemfile because it's a _default_ gem. Which do you think is better? Feel free to make either change to this PR or defer to a future PR since it's just a warning.

Note that we have to add the dependency on bigdecimal because it's a _bundled_ gem, which means [it's installed with ruby but it can be removed by users](https://stdgems.org/). If users removed it, then our dependency is missing and this library breaks. So we have define it in the gemspec.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         ruby:
+          - "3.4"
           - "3.3"
           - "3.2"
           - "3.1"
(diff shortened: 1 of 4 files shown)


timriley:
Thanks @cllns, this looks good to me. I'll merge this one now and I reckon we can just add `"ostruct"` to the `Gemfile` in a follow-up PR.

PR #419, opened 2026-07-01
Allow bigdecimal 4 or 3

sandbergja:
It would be nice to be able to use the latest and greatest `bigdecimal` with Hanami.  All of the tests around bigdecimal pass for me with bigdecimal 4


timriley:
Thanks @sandbergja! I've just released this as 3.0.1 :)

